PR #7, opened 2015-06-23
Return React Elements from formatData

### LINKED ISSUE #6: React Component for Column/Data Format
It would be useful to be able to use React Elements in the `dataFormat` prop.

aaronhayes:
Fixes #6


aaronhayes:
There's a typo in this PR. won't fix issue. Closing PR. Will create new one with proper fix soon.

PR #12, opened 2015-07-10
removed the react TestUtils addons in the TableBody

sridharAJ:
reactjs is not shipping the TestUtils addon in the production environment, In the TableBody.js you are using isElement from TestUtils addons to check the formattedvalue. because of that its breaking in the production. so i have removed that TestUtils addon function. And used the React.isValidElement function.


AllenFang:
@sridharAJ, I've merge your PR and release v0.8.7. Thanks for your contribution :)

PR #18, opened 2015-07-26
Added 'selected' as a prop to allow setting selected state externally

low-ghost:
Needed to be able to save checked item list further up so that it could be passed in again if the component was unmounted and then remounted (via react-router in this case). Simple changes just to allow 'selected' array as a prop.

The super call changes were just due to babel yelling at me that you cannot access 'this' before a super call. You can safely ignore them if this was just oddities in my transpiling process.

Also, on an unrelated note, setting props to true with prop={true} is unnecessary. It can certainly be clearer, but I think this looks a bit cleaner:

```js
<BootstrapTable data={products} striped hover pagination search>...
```
Anyhow, thanks for the great work on this.

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -108,6 +108,7 @@ class BootstrapTable extends React.Component{
             keyField={this.store.getKeyField()}
             condensed={this.props.condensed}
             selectRow={this.props.selectRow}
+            selected={this.props.selected}
             cellEdit={this.props.cellEdit}/>
           {tableFilter}
         </div>
@@ -315,7 +316,8 @@ BootstrapTable.propTypes = {
   insertRow: React.PropTypes.bool,
   deleteRow: React.PropTypes.bool,
   search: React.PropTypes.bool,
-  columnFilter: React.PropTypes.bool
+  columnFilter: React.PropTypes.bool,
+  selected: React.PropTypes.array
 };
 BootstrapTable.defaultProps = {
   height: "100%",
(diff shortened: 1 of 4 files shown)


AllenFang:
Hi @low-ghost , your contribution is very well and it's a nice feature for react-bootstrap-table. After review your PR, I think I'll remove the <code>selected</code> prop from <code>&lt;BootstrapTable&gt;</code> and take it as an attribute in <code>selectRow</code> prop on <code>&lt;BootstrapTable&gt;</code>, as follow:
<pre>
var selectRowProp = {
  mode: "checkbox",
  clickToSelect: true,
  selected: [1,5,10......],
  bgColor: "rgb(238, 193, 213)",
  onSelect: onRowSelect,
  onSelectAll: onSelectAll
};
//...
&lt;BootstrapTable data={products} selectRow={selectRowProp}....
</pre>

And we need to verify that user choose the single select so they can't give multiple values in <code>selected</code>

Anyway, I'll complete it and thanks for your contributions, you did a great job!! :)

AllenFang:
I've merge your PR and do some modified like I said. The newest version is v0.8.8 :)

low-ghost:
Great, thanks! Working perfectly, and definitely a good idea to put the prop on the selectRow object. Thanks for the quick response and glad I could lend a hand.

stephenrs:
Really nice work guys! Thanks for this.

AllenFang:
There's some bug about this fix, so please update to v0.9.0 :)

PR #32, opened 2015-08-15
Show a message when the table is empty

Julusian:
Currently, when tables have no data, or a search returns no data the table looks a little odd and empty. This PR adds the below behaviour instead of rendering the table with no rows.

screenshot from 2015-08-15 10-38-27


AllenFang:
Hi @Julusian, thanks for your contributions :)

PR #47, opened 2015-09-04
Fix react 'missing key' warning

sirreal:
\<TableHeader/\> was creating children via a loop
with no `key` property set. This causes a warning in React.

Added key (index) to the new children created.


AllenFang:
@sirreal , Thanks for your contribution :)

sirreal:
My pleasure, and thanks for your work. I'm using this for a project right now and if I have some time I'll try to submit some more PRs with some of the work I've done.

PR #61, opened 2015-09-28
Fix "window is undefined" bug on node.js

tbo:
Node.Js will throw an error if you check for "window" directly. This is a workaround for that problem.


AllenFang:
Thanks :)

Issue #70, opened 2015-10-12
Not Preserving Pagination and Ordering after setData


ben-ward:
I'm using the BootstrapTable with the following options.

<BootstrapTable search={true} data={this.dataSet} pagination={true} striped={true} hover={true} className="table table-striped table-bordered nowrap" width="100%">

After I use setData the current page and any ordering is lost. 

I could rerender the component if I had some way of passing the current page and ordering field via the options attribute. Is there anyway to get the current page or ordering?

Great work BTW. This was really helpful.

Thanks,
Ben

02ps719:
I have fixed this. Shall send in a pull request soon. Also fixed the an issue with search filter not filtering when setData was being used to add/edit/delete rows matching the search criteria.

ben-ward:
@02ps719 nice! that would be great

AllenFang:
Thanks @02ps719's work :)  :+1:  and about @ben-ward's said, I'll think about how to get curr page and ordering, but if I give you some hook for page changing or data sorting and pass data which you want, maybe it's a good way to this enhancement.

AllenFang:
HI all, this issue has been solved on v1.2.1 :)

AllenFang:
About how to get current page and get ordering, can use the newest version and see <a href="https://github.com/AllenFang/react-bootstrap-table/blob/master/CHANGELOG.md">CHANGELOG.md</a>

PR #77, opened 2015-10-16
Ensuring pagination is maintained when using TableDataSet as the data model instead of a simple array

02ps719:
There was an issue wherein when using TableDataSet as the model, calling setState on add/edit/delete on the model was not paginating correctly(as compared to Add/Delete/Edit row show in the demo - which uses a simple array as the model)

Am more inclined to think to have an enum passed to setState method to know what exactly is the state change - add/edit/delete and handle pagination accordingly than to rely on the using the old data length as I have used. 

@AllenFang , my machine's python and node is failing, so could not run gulp to regenerate the bundle nor run the tests. Can you please run that before merging the pull request?

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -37,7 +37,7 @@ class BootstrapTable extends React.Component{
 	  this.props.oldDataLength = _data.length;
 	  this.props.data.clear();
       this.props.data.on('change', (data) => {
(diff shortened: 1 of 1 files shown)


AllenFang:
HI @02ps719, I think the real problem should be in page changing and data sorting, not in constructor. Anyway, I'll review your PR and I'm sure glad of your contributions. :+1:

AllenFang:
HI @02ps719, I've fix this issue about #70, you can check the newest version of code. Anyways, thanks your contributions :)

PR #95, opened 2015-10-22
updated dependencies to work with node 4.2.1

kirstein:
`contextify` build fails on node 4.2.1. This fixes that.


AllenFang:
@kirstein , why drop this PR?

PR #120, opened 2015-11-05
Feature - export to csv

khinlatt:
Enable csv export link in the toolbar.

This uses FileSaver.js ( https://github.com/eligrey/FileSaver.js/ ) to write the filteredData to proper tab delimited rows.

Usage:
```
pass in exportCSV={true} to BootstrapTable component to enable the link in the toolbar.
```


AllenFang:
@khinlatt, I will review your codes, Anyway, thanks a lots!

PR #122, opened 2015-11-06
Feature (add callback hook to get store's data) - after search filter hook

khinlatt:


khinlatt:
@AllenFang sorry wrong repo again. This is still WIP.

Issue #113, opened 2015-11-03
How to tap into filteredData that is being display?


khinlatt:
@AllenFang, is there an easy way to expose the filteredData (after searching) back out to the component, rendering BootstrapTable?

AllenFang:
@khinlatt, maybe I can support the after filtering and searching hook and pass the filtered data as params

khinlatt:
@AllenFang actually I am just going to make another PR instead. I think it will be a good feature to this component. It will print out whatever is on the table to CSV. Going to close this issue. Thanks.

AllenFang:
OK, any problems about api design, you can ask me :)

khinlatt:
great. will do. appreciate it.

khinlatt:
@AllenFang hey when you say "pass the filtered data as params", could you explain a bit more? It would be of great help if I can access to the filteredData from my react components somehow.

AllenFang:
I mean the <code>react-bootstrap-table</code> will take the filtered data as params and pass to the hook function, if I implemented, it will look like following:

```javascript
var options = {
    afterFilitering:  function(filteredData){
         //filteredData is data which apply your filter expression
    },
    afterSearch:  function(result){
        //result is searched result
    }
}
//..
<BootstrapTable options={options} ...>
```

khinlatt:
Ah I see. That's pretty neat. I do have a follow-up question. When you also turn on pagination, filteredData will always be whatever on the first page. Correct? How can we also tap into all of the filteredData (from Search), not just the one on the first page.

"Export to CSV" should really be producing all the filteredData.

Thanks a lot.

AllenFang:
It's correct, if you use the <code>store.get()</code> and enable pagination, the result will be the data which in current page, but you explain is right, I should support another function on <code>store</code> for get the all result(I'll do it in tomorrow). However, the after searching and filtering hook, did you implement? if not, I can do it in tomorrow :)

khinlatt:
If you can look into that CSV PR and work on that store to get all the result. That would be great.

I will attempt at searching and filtering hook. Will keep you posted. Are searching and filtering actually two different things? Can you explain me the difference between those two? Thanks.

AllenFang:
Searching and filtering is two different thing, so I will separate <code>afterSearchOrFilter</code> to <code>afterSearch</code> and <code>afterFilter</code> base on your PR #123

AllenFang:
[...]

PR #123, opened 2015-11-06
Feature - add search filter hook

khinlatt:
Added a callback function hook so any component rendering ReactBootstrapTable could have access to the filteredData.

Usage:

```
handlerAfterSearchOrFilter(filteredData) {

}

<ReactBootstrapData options={{afterSearchOrFilter: this.handleAfterSearchOrFilter}}>
```

NOTE***: I did not check in the updated files in lib and dist; so need to run build before pulling in.


src/store/TableDataStore.js
```diff
@@ -209,6 +209,11 @@ export class TableDataStore {
     }
   }
 
+  getDataIgnoringPagination() {
```
khinlatt: @AllenFang exportCSV could just hit this.store.getDataIgnoringPagination() instead of this.store.get() to get all the results.

AllenFang: Perfect, calling function <code>getCurrentDisplayData</code> is only correct soultion

AllenFang:
HI @khinlatt, why call <code>afterSearchOrFilter</code> function on componentDidUpdate in <code>BootstrapTable</code>. If user disable search or column filter but assign a <code>afterSearchOrFilter</code>, when something event trigger componentDidUpdate, the <code>afterSearchOrFilter</code> will be called. So I think we need separate search and column filter like I said <a href="https://github.com/AllenFang/react-bootstrap-table/issues/113#issuecomment-154625549">#113</a>, and then put the the two hooks function on <code>handleSearch</code> and <code>handleFilterData</code>(also in BootstrapTable.js) should be better

khinlatt:
@AllenFang I see. Thanks for the help and fast response!

AllenFang:
You're welcome, also thanks your helpful contributions :)

PR #130, opened 2015-11-09
Bug (csv export) - bug on csv export

khinlatt:
- need to ignore hidden columns (what's shown is what's printed out)
- ignore pagination


AllenFang:
Oh! I certainly forgot to use <code>getDataIgnoringPagination</code>.  Thanks you to keep track this function and did more improvement :)

PR #139, opened 2015-11-13
Bug (multiColumnSearch) - pass multiColumnSearch

khinlatt:
@AllenFang when the store's setData is called, we should pass in multiColumnSearch param as well. If not, when that happens, we lose that feature in the store.

I do have a follow up question for you. When is setData change event triggered?


AllenFang:
Hi @khinlatt, thanks your contribution, answer your question first, setData function will be called in <code>constructor</code> and <code>componentWillReceiveProps</code> in <code>BootstrapTable</code>. The purpose of this function is to re-assign the data to store, just like re-assign data to table. Anyway, I thinks the better solutions is not pass the <code> multiColumnSearch </code> in the <code>setData</code> maybe <code>setProps</code> or something like that, so I'll create an issue for fix this bug and do some refactoring. Thanks a lots :)

khinlatt:
I see. Please let me know after you've done the refactoring. Thanks!

AllenFang:
OK :)

Issue #141, opened 2015-11-14
Missing multiColumnSearch if data reload


AllenFang:
#139

PR #142, opened 2015-11-15
Fix Missing multiColumnSearch if data reload

AllenFang:
#141 #139 @khinlatt


khinlatt:
i see. thanks! @AllenFang

PR #143, opened 2015-11-16
Chore: csv back to comma delimited

khinlatt:
@AllenFang excel default open is by comma delimited so it's annoying to open manually with tab delimited.

So I changed it back to comma delimiter with "" wrapped around to escape the commas in cell.


AllenFang:
Great job:) Thanks

kintel:
Note: Excel csv format depends on locale. Semicolon as a separator is very common, at least in Europe. We should probably externalize the whole csv function to a library which can configure the output better.

Issue #149, opened 2015-11-23
Server side rendering and filesaver.js


staminaloops:
Hi @AllenFang 
I'm using SSR on my project, and had some problems with filesaver.js

I fixed it in my fork, it will only load filesaver on the browser bundle.
Check it here: https://github.com/staminaloops/react-bootstrap-table/commit/e4c134a9e20dd7ebeaefa82dc3292634c3816d88#diff-3602ea699cf9dc846a0e23263aecf3b0

If you agree, I can do a pull request.

Cheers :)

AllenFang:
@staminaloops, you did a great job, thanks a lots. Feel free to make PR :)

PR #150, opened 2015-11-23
Only load filesaver.js on the client

staminaloops:
Fix https://github.com/AllenFang/react-bootstrap-table/issues/149


AllenFang:
Thanks your contribution :+1:

PR #153, opened 2015-11-25
don't draw pagination dropdown for <=1 options

frontsideair:
If there is one or zero options for pagination size list, don't draw a dropdown


AllenFang:
Good enhancement :) Thanks :+1:

frontsideair:
You're welcome!

PR #158, opened 2015-11-30
TableDataSet is now available via Window object

Agilesolutionspk:
Example use
var ReactBsTable = window.BootstrapTable;
var ReactTds = window.TableDataSet; //import TableDataSet
var dataSet = new ReactTds(products);


AllenFang:
Thanks for this bug fixing :)

PR #159, opened 2015-11-30
add a select class on selection of a row

LensArt:
add an additional class if a row is selected. This can be used io the background color.

### DIFF
--- a/lib/TableRow.js
+++ b/lib/TableRow.js
@@ -49,7 +49,7 @@ var TableRow = (function (_React$Component) {
         style: {
           backgroundColor: this.props.isSelected ? this.props.selectRow.bgColor : null
         },
-        className: this.props.className || ''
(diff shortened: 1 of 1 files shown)


AllenFang:
@LensArt, Thanks your contributions :) But I think you should patch the source code for me.

LensArt:
@AllenFang Ok I am not sure what you mean? You want this kind of patch? https://patch-diff.githubusercontent.com/raw/AllenFang/react-bootstrap-table/pull/159.patch
Or what should I provide you with?

AllenFang:
@LensArt, you should modify the <code>./src/TableRow.js</code> instead of <code>./lib/TableRow.js</code>. Because under <code>lib</code> folder is all the bundled files not the source code

LensArt:
Oops :# I fixed it in another PR :)

PR #178, opened 2015-12-11
Use onSelect handler return value for checking/unchecking

frontsideair:
If onSelect handler function returns false, don't check or uncheck the checkbox, for any other value behave normally.

This is required to handle and use selection state outside the component.


frontsideair:
There seems to be a version problem with react-router and history. I'll also add documentation.

frontsideair:
I made another commit for onSelectAll return values and documented the change in gh-pages branch, @AllenFang. I can open a PR as soon as the version problem is resolved.

AllenFang:
@frontsideair, sorry, I can't got your certainly problem, what the main reason for should return bool value on <code>onSelect</code>, because <code>onSelect</code> will take <code>row</code> and <code>isSelected</code> as arguments for you, so I think it can use selection state outside the component (because you know which row is selected or unselected). hope I don't misunderstand your question ;(  my english is bad ;(

frontsideair:
@AllenFang Well in my case, the user needs to select two rows to compare them, and selecting more than two should be illegal. When I was doing this with a plain table, I made it so a row wouldn't be selected if there already were two selected rows. I couldn't do it with this component, so I thought checking return value of handler would be a good solution. Of course there might be a better solution that I didn't think of, I'd like to hear it if you have a suggestion.

AllenFang:
@frontsideair, ok I understand :), your solution is best to do that, because you can handle the selection result on component right now, don't through action to store to valid the restrict for two row selection.

Thanks your explantation and did you want to make PR? if you have no time, I can complete this at next version :)

frontsideair:
I just did!

PR #180, opened 2015-12-12
Handle return value from onSelect and onSelectAll handlers

frontsideair:
If the return value is equal to false, don't select or unselect; for any other return value behave normally.

Sorry for including production files, I'll omit them next time.


AllenFang:
Good job! Thanks :)

frontsideair:
You're welcome! There's more to come from me.

PR #182, opened 2015-12-13
fix a case where current page and pagination size is lost

frontsideair:
This problem is hard to explain, so I'll provide a use case along with an explanation here: https://gist.github.com/frontsideair/2ec7c1356b4001366ef9

Here I handle the selection state outside the component so I can control the number of rows. But when a row is selected or deselected, this triggers a rerender. Upon a rerender, table asks the store for visible data and when doing that checks if the page is changed. But since in this case a rerender was triggered without changing the page, it renders with default pagination size and page number.

I'm not sure if my solution is problem-free, but if this.refs.pagination is available, I think it's safe to assume there's a pagination component available.


AllenFang:
@frontsideair, I'll close this PR because I used a better solution for this bugs, you can check <code>master</code> branch for the both commits [64ed6d3](https://github.com/AllenFang/react-bootstrap-table/commit/f01f6ec2d9900bf7a11c5a6413b61c177c26e701) and [f01f6ec](https://github.com/AllenFang/react-bootstrap-table/commit/f17e4a4b85b924f8ffab1a9479d8789fb8dc8239). Anyway, I'll release in next version.

But your another PR I'll merge in these day, thanks your contributions :ship:

frontsideair:
Looks like I made the PR wrong, I thought isolated my commit to another branch but sent the PR from master, which is weird.

I'll wait for your solution, thanks.

AllenFang:
yes, make a branch by each issue will be better :)

PR #183, opened 2015-12-13
Style improvements for pagination and toolbar

frontsideair:
I used react-bootstrap elements where I could and tried to use whitespace better.

### DIFF
--- a/css/react-bootstrap-table.css
+++ b/css/react-bootstrap-table.css
@@ -54,9 +54,6 @@
   width: 100%;
 }
 
-.react-bs-container .tool-bar{
-  margin-bottom: 5px
-}
 /*inline editor default style*/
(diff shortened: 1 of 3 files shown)


AllenFang:
@frontsideair, <code>react-bootstrap-table</code> doesn't depended on <code>react-bootstrap</code>,  it only depends on <code>dev</code>. So can you provide another solution for it? However, I don't want to depend <code>react-bootstrap</code> with <code>react-bootstrap-table</code>.

frontsideair:
Sorry, I assumed react-bootstrap was a production dependency. I'll take a look at it as soon as possible.

AllenFang:
that's ok, take your time :)  Thanks very muck :+1:

PR #184, opened 2015-12-14
Don't hide pagination size selector column when there isn't a pagination dropdown

frontsideair:
It broke the pagination buttons.


AllenFang:
@frontsideair, thanks :)

PR #193, opened 2015-12-29
Make style improvements

frontsideair:
Sorry this took so long but I made my style improvements not depend on react-bootstrap, you can review and merge if you want.

### DIFF
--- a/src/toolbar/ToolBar.js
+++ b/src/toolbar/ToolBar.js
@@ -119,7 +119,9 @@ class ToolBar extends React.Component{
             <i className="glyphicon glyphicon-trash"></i> Delete
           </button>:null;
     var searchTextInput = this.props.enableSearch?
-      <input type='text' placeholder={this.props.searchPlaceholder?this.props.searchPlaceholder:'Search'} onKeyUp={this.handleKeyUp.bind(this)}/>:null;
+      <div className="form-group form-group-sm">
+        <input type='text' placeholder={this.props.searchPlaceholder?this.props.searchPlaceholder:'Search'} onKeyUp={this.handleKeyUp.bind(this)}/>
+      </div>:null;
     var modal = this.props.enableInsert?this.renderInsertRowModal(modalClassName):null;
     var warningStyle = {
       display: "none",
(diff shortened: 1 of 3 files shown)


AllenFang:
hi @frontsideair, I'll review it. thanks u very much  :+1:

frontsideair:
I added one more commit to fix some bugs.

src/toolbar/ToolBar.js
```diff
@@ -119,7 +119,9 @@ class ToolBar extends React.Component{
             <i className="glyphicon glyphicon-trash"></i> Delete
           </button>:null;
     var searchTextInput = this.props.enableSearch?
-      <input type='text' placeholder={this.props.searchPlaceholder?this.props.searchPlaceholder:'Search'} onKeyUp={this.handleKeyUp.bind(this)}/>:null;
+      <div className="form-group form-group-sm">
+        <input className="form-control" type='text' placeholder={this.props.searchPlaceholder?this.props.searchPlaceholder:'Search'} onKeyUp={this.handleKeyUp.bind(this)}/>
```
imaireaux: Hi,

I am currently searching for a way to customize some elements like the search input.
I can see that you add the `form-control` class to it.
May be you could make it customizable in this PR ?

Something like : 
`<input className={this.props.searchClassName?this.props.searchClassName:'form-control'} type='text' placeholder={this.props.searchPlaceholder?this.props.searchPlaceholder:'Search'} onKeyUp={this.handleKeyUp.bind(this)}/>`

AllenFang: @imaireaux, id you want to use default search element in <code>react-bootstrap-table</code>, you can disable <code>search</code> and customize yours and call <code>handleSearch</code> on <code>BootstrapTable</code>, check the example on following:
```js

function search(text) {
    this.refs.table.handleSearch(text);
}

<BootstrapTable ref='table' data={products}>
    //...
</BootstrapTable>
```
See the <code>handleSearch</code> [docs](http://allenfang.github.io/react-bootstrap-table/docs.html)

AllenFang:
Nice improvement :)  I'll merge it on next version. Thanks

AllenFang:
Released on <code>v1.4.0</code>, thanks again:)

frontsideair:
You're very welcome

src/toolbar/ToolBar.js
frontsideair: @imaireaux I understand your expectation but I'm afraid it'd bloat the props if we passed this class name to the component. The more general solution to your problem would be to add a class to the main container and write a CSS selector to affect search input specifically. Something like this:

```
.react-bs-container.my-table .tool-bar .form-group input {
  color: red;
}
```

imaireaux: You're both right. Before this PR, this input had no bootstrap class, but this is not the case anymore.
And it's true there are other ways to customize this input.

frontsideair: You're welcome.

PR #195, opened 2016-01-01
Add property to disable confirm delete row and allow custom confirmation handler

SebastianCoetzee:
Added a boolean property to `BootstrapTable` called `confirmDeleteRow` that is set to `true` by default. If the property is set to `false`, the user will not be prompted to confirm when deleting a row.

Added a function property to `BootstrapTable` called `handleConfirmDeleteRow` which allows the user to provide a custom handler for the confirmation prompt when deleting a record. The `handleConfirmDeleteRow` function receives a callback function as parameter that should be called without parameters if the deletion is confirmed.

Here is an example of using the `handleConfirmDeleteRow` property:

``` javascript
var React = require("react");
var ReactDOM = require('react-dom');
var ReactBsTable  = require('react-bootstrap-table');
var BootstrapTable = ReactBsTable.BootstrapTable;
var TableHeaderColumn = ReactBsTable.TableHeaderColumn;

module.exports = React.createClass(
	{

		getInitialState: function(){

			return {
				data: [
					{
						id: 1,
						name: "John Smith",
						email: "john@smith.com",
						created: new Date()
					},
					{
						id: 2,
						name: "Jane Smith",
						email: "jane@smith.com",
						created: new Date()
					}
				]
			};

		},

		//This is an example of a custom confirm function.
		customConfirm: function(next){

			if (confirm("Are you sure you want to delete?")){

				//If the confirmation is true, call the function that
				//continues the deletion of the record.
				next();
				
			}

		},

		render: function() {

			return (
				<BootstrapTable
					data={this.state.data}
					striped={true}
					hover={true}
					condensed={false}
					pagination={true}
					search={true}
					keyField={"id"}
					deleteRow={true}
					confirmDeleteRow={true}
					handleConfirmDeleteRow={this.customConfirm}
					selectRow={
						{
							mode: "radio"
						}
					}>
					<TableHeaderColumn dataField="id" hidden={true} >ID</TableHeaderColumn>
					<TableHeaderColumn dataField="name" dataSort={true}>Name</TableHeaderColumn>
					<TableHeaderColumn dataField="email">Email</TableHeaderColumn>
					<TableHeaderColumn dataField="created">Created</TableHeaderColumn>
				</BootstrapTable>
			);

		}
	}
);
```


AllenFang:
@SebastianCoetzee, thanks for your contributions :+1:  I'll review your code and release it on next version :ship:

AllenFang:
@SebastianCoetzee, I'll cancel <code>confirmDeleteRow</code> props but remain <code>handleConfirmDeleteRow</code>. if <code>handleConfirmDeleteRow</code> given, <code>react-bootstrap-table</code> will execute the given function to process confirmation

AllenFang:
Released on <code>v1.4.0</code>, thanks!

SebastianCoetzee:
No problem! :)

PR #199, opened 2016-01-04
Fix rare bug where pagination lost

frontsideair:
After updating I found out I still had problems with pagination size and page getting lost. You can recreate the problem in the updated gist: https://gist.github.com/frontsideair/2ec7c1356b4001366ef9

I think the code is a little bit simplified while covering all the cases previously covered. I just made it so the priority of pagination selector is higher than props. If you see any case missing just let me know. I tested it by hand so it may be a good idea to test it yourself.

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -78,22 +78,14 @@ class BootstrapTable extends React.Component {
   getTableData(props) {
     let result = [];
     if (props.pagination) {
-      let page = props.options.page;
-      let sizePerPage = props.options.sizePerPage;
-      if(!page){
-        if(!this.refs.pagination){
-          page = 1;
-        } else {
-          page = this.refs.pagination.getCurrentPage();
-        }
+      let page = props.options.page || 1;
+      let sizePerPage = props.options.sizePerPage || Const.SIZE_PER_PAGE;
+
+      if(this.refs.pagination) {
+        sizePerPage = this.refs.pagination.getSizePerPage();
+        page = this.refs.pagination.getCurrentPage();
       }
-      if(!sizePerPage){
-        if(!this.refs.pagination){
-          sizePerPage = Const.SIZE_PER_PAGE;
-        } else {
-          sizePerPage = this.refs.pagination.getSizePerPage();
-        }
-      }
+
       // #125
       if(page >= props.data.length / sizePerPage) page = 1;
       result = this.store.page(page, sizePerPage).get();


AllenFang:
Thanks @frontsideair, I'll spend some time to review this PR :).  Good job!

AllenFang:
@frontsideair, sorry, I made a big mistake, I didn't merge both commits [64ed6d3](https://github.com/AllenFang/react-bootstrap-table/commit/f01f6ec2d9900bf7a11c5a6413b61c177c26e701) and [f01f6ec](https://github.com/AllenFang/react-bootstrap-table/commit/f17e4a4b85b924f8ffab1a9479d8789fb8dc8239) into master. I'll fix it on next version. Anyway, this condition is same as [#182](https://github.com/AllenFang/react-bootstrap-table/pull/182), so I'll close this PR and patch these two commits

AllenFang:
@frontsideair, I've push patch to <code>master</code>, you can install <code>react-bootstrap-table</code> by github format or clone it to test your scenario

frontsideair:
I pulled the project and tried it both on my actual project and the test case I provided; it seems like problem persists.

You can review my PR again, but as far as I can tell from my debugging the problem arises from the logic preferring default pagination size from props over checking the current state (with `this.refs.pagination.getSizePerPage()` call). All I did was overwrite default pagination size from props with the one from state if a ref to pagination exists. This cleared a lot of `if`s as a side effect too.

AllenFang:
but I test you case, the pagination does not lost anymore. anyway, I'll try again and check this PR. Thanks :+1:

AllenFang:
@frontsideair, I check your branch <code>pagination-lost</code>. I think you need to merge the newest code again(I think we based on different code base), because I check out your PR but the problem still remain. So could you please check it again?

AllenFang:
@frontsideair, can you try the newest version again :)

bastiao:
@AllenFang this branch really fixes the problem. 1.4.2 does not fixes the problem. Did you consider to merge?

AllenFang:
@bastiao, thanks, I'll try the [gist](https://gist.github.com/frontsideair/2ec7c1356b4001366ef9) again to find the problem. Because I can't reproduce this issues according to this gist, so any help debugging your situation would be appreciated :)

bastiao:
@AllenFang I tried to start developing process to fix issue but it raises me:

Uncaught TypeError: Cannot read property 'ToastContainer' of undefined: Notification.js:31

any idea why?

bastiao:
Thanks for merging. Now I'm trying to handle the pagination but it is tricky because you changed a bit since 1.4.0

AllenFang:
@bastiao, thanks a lots, feel free to ask anything or join the [gitter](https://gitter.im/AllenFang/react-bootstrap-table) to discuss

frontsideair:
Sorry for being unresponsive past couple of days. I guess I did provide the code, but missed the steps to reproduce the bug.

When running the code I provided in the latest version, I would change pagination size from default value to something else, then change the page or check a checkbox, and see that the pagination size reverted to the default value. (But it still shows the value I picked.)

bastiao:
@frontsideair you branch is a bit out of dated. Did you have fix for 1.4.2?

frontsideair:
I can update my branch to 1.4.2 today I guess.

bastiao:
@frontsideair it will be good, but there are conflits there.

frontsideair:
It looks like the code changed a bit so I had to find another solution, please check this branch: https://github.com/frontsideair/react-bootstrap-table/tree/pagination2 It is up to date with master.

bastiao:
@frontsideair I did it, but I do not believe that it solves the problem. It is quite strange.

bastiao:
Ok, found the problem. TableColumn (and/or) TableRow use the same id. Thus, data-reactid is the same and it does not render. 
This branches fixes the issue (but with hacks, not in general case). Do you know what should we pass for key?
{i} is not a solution. 
https://github.com/bastiao/react-bootstrap-table/tree/new/prod2

AllenFang:
@bastiao, did you really solved this issue?

frontsideair:
I'll go ahead and close this PR, since it's out of date with `master` and I have another branch. Also the issue @bastiao mentioned seems to have nothing with the one I was trying to solve. I'm opening another PR with my up-to-date branch.

AllenFang:
Thanks @frontsideair

PR #216, opened 2016-01-18
Fix doc assets being incorrectly served over http

rogerhutchings:
Some of the assets on the docs branch aren't being rendered (certainly in Chrome, probably in other browsers) as they're being served over `http` rather `https`, which breaks some of the examples.


AllenFang:
@rogerhutchings, thanks you

PR #218, opened 2016-01-19
Fixed issue with ToastContainer.

bastiao:
It does not run without that. 

```
Uncaught TypeError: Cannot read property 'ToastContainer' of undefined: Notification.js:31
```

Just did: 
```
npm install 
gulp example-server
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   },
   "dependencies": {
     "classnames": "^2.1.2",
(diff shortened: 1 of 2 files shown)


Enet4:
:+1: As it was, Notification.js was not even valid ES2015. Definitely a good idea to merge.

AllenFang:
Thanks :+1:  BTW, why I have ever seen this problem ;(

PR #228, opened 2016-01-22
Fix pagination lost bug

frontsideair:
I've changed the order of preference for page and pagination size when receiving props. It would prefer nextProps over current page, which caused problems for me. It might work when you handle the selection inside the component, but as I'm giving this as a prop, it's triggered at times when it wasn't expected.

With this PR it'll prefer current page and pagination size over the ones provided by props.

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -119,10 +119,8 @@ class BootstrapTable extends React.Component {
     if (Array.isArray(nextProps.data)) {
       this.store.setData(nextProps.data);
       let paginationDom = this.refs.pagination;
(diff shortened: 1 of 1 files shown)


AllenFang:
@frontsideair, thanks, I'll review it :+1:

AllenFang:
@frontsideair, could you give me a example for showing this issues, I want to add and explain it on ```CHANGELOG```

frontsideair:
How about this? https://github.com/AllenFang/react-bootstrap-table/pull/233 It's almost the same as the gist I shared to test my error case, it's also illustrative for managing state outside the component.

PR #232, opened 2016-01-23
Fix #197

### LINKED ISSUE #197: bootstrap.js dependency
I noticed that we need `bootstrap.js` as a global runtime dependency as the pagination size dropdown functionality depends on a `data-toggle` property which is provided by Bootstrap JS code. I'm considering moving that functionality to React as a way to remove this global dependency.

If there is anyplace else using `bootstrap.js` that you can point out, I'll keep that in mind when doing that.

frontsideair:
Removes jQuery and `bootstrap.js` dependencies, but adds `react-onclickoutside` (which is a good tradeoff). I removed those scripts from `examples/index.html` for good measure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "react-bootstrap": "^0.27.2",
     "react-dom": "^0.14.3",
     "react-hot-loader": "^1.3.0",
+    "react-onclickoutside": "^4.5.0",
     "react-router": "^1.0.2",
     "react-router-bootstrap": "^0.19.2",
     "style-loader": "^0.13.0",
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "react-bootstrap": "^0.27.2",
     "react-dom": "^0.14.3",
     "react-hot-loader": "^1.3.0",
-    "react-onclickoutside": "^4.5.0",
     "react-router": "^1.0.2",
     "react-router-bootstrap": "^0.19.2",
     "style-loader": "^0.13.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -48,6 +48,7 @@
     "react-bootstrap": "^0.27.2",
     "react-dom": "^0.14.3",
     "react-hot-loader": "^1.3.0",
+    "react-onclickoutside": "^4.5.0",
```
staminaloops: maybe it should be a dependency and not a dev-dependency?

frontsideair: I don't know. I thought since it's small enough, we don't need to increase the number of runtime dependencies and could just bundle it. I can change it if you think it makes a difference.

AllenFang: I thinks it should be a dependency :)

frontsideair: Sure, I'll change it to da dependency while I'm on it.

frontsideair:
Should be okay now, please review.

AllenFang:
@frontsideair, I found not only dropdown menu for pagination but also insert row for popup a modal will also use ```bootstrap.js```. sorry, this is my fault, could you please also fix the insert button for your solution. Because I need to release today, so I revert this PR and will merge it in next version. Sorry

frontsideair:
No problem, I'll take a look as soon as possible. If there's anyplace else using `bootstrap.js` you can add a comment and I'll check them too.

AllenFang:
@frontsideair, take your time, currently I just know the insert button for modal and the close button in modal. Thanks :)

DongShi:
Hello there, by now, this library still depends on bootstrap.js and jquery right?
I just port one demo code to my own project and found that the dropdown list for pagination did not work.

PR #242, opened 2016-01-28
enable debugging original jsx files in while running example-server

dana2208:


AllenFang:
@dana2208, what's the formatExtraData?

dana2208:
I mistaken, I wanted to make a different pull request but it was added to the previous pull request, sorry!
The formatExtraData enbables to set data that is neither in the row nor the value of the cell and that is needed for formatting.
For example I want to format a value that comes from an enum:
`type: { male: 0, female: 1}`
`status: { married: 0, single: 1}`
but when formatting I have no idea what the name of the original enum is, I only have the value: for example 0 so I don't know whether to display "male" or "married". I need as extraData the name of the enum (either "type" or "status").
This may be good also in many other scenarios, and formatExtraData can be anything the user wants to pass.

AllenFang:
@dana2208, that's fine, you should implement these two features in different branch so that you can open two PR separated. Anyway, you idea is awesome, but could you please provide an example under ```examples/js/column-format``` folder? you can just commit on this branch, don't separate :ship: 

Thanks your contribution

PR #248, opened 2016-01-31
add an example for extra data formatting

dana2208:
@AllenFang I see that in 1.4.6 you added an example, maybe you'll like more my example so let me know what you decide.
Moreover I found an issue when inserting a new row that has a formatted column.
I fixed it also in this pull request.
The issue is in BootstrapTable.js in `renderToolBar()` line 555:
`format: props.format ? format : false;`
Neither `props.format` nor `format` exist.
The right code is:
`format: props.dataFormat ? function(value){
              return props.dataFormat(value, null, props.formatExtraData).replace(/<.*?>/g,'');
            } : false`
like in TableBody.js.
I passed null as the second parameter, since the signature of this function is:
argument 1: value
argument 2: the data of the current row
argument 3: any extra data needed for the formatting
The first and second arguments make sense when editing the cell, but when inserting a new row they are useless and therefore we can pass null in the second parameter. Indeed there is no sense in trying to get the current row values when this current row is beeing created.

If you want to reproduce the issue, take the insert-type-table.js example (in advance folder) and add a formatter in one of the column.
You will see that when opening the modal for inserting a new row, the field is not formatted.

Thanks


AllenFang:
thanks :) I'll check this out !

AllenFang:
@dana2208, thanks your contribution, I've released on v1.5.0 :+1:

PR #253, opened 2016-02-02
Show default caret to tell user which column is sortable

arowM:
Some people can not notice that they can sort the table column by clicking the header cell or which column is sortable when they simply see the sortable table at first.
I suggest to show default caret on sortable column to enhance the value of this awesome library.

sort
sorting

Thanks


AllenFang:
@arowM, Nice patch :)

arowM:
Thanks!

AllenFang:
@arowM , thanks your contribution, I've released on v1.5.0 :+1:

PR #256, opened 2016-02-05
Change pagination behavior and add more customization options to it

diazproduction:
Hi, I made some changes and adjustments to the way in which the pagination of the tables works:

- If the table has only one page (or no pages at all), it won't show the navigation controls (first, prev, next, last).
```
[ 1 ]
```
- If the user is on the very first or the very last page of the table, it won't show the first/prev or next/last buttons respectively:
```
([ 1 ]) [ 2 ] [ 3 ] [ 4 ] [>] [>>]

[<<] [<] [ 7 ] [ 8 ] ([ 9 ])
```
- If the first or the last page is within the range of the pagination size, it won't show the first or last button respectively:
```
[<] [ 1 ] ([ 2 ]) [ 3 ] [ 4 ] [>] [>>]

[<<] [<] [ 7 ] ([ 8 ]) [ 9 ] [>]
```
(_These two last changes were made because in that way you can instantly show to the user where the table begins and ends._)

I also add some props on the _options_ in order to be able of changing the text of the navigation controls, by default they are as they were before (<< < > >>) but now you can specify the name you want to give to it in case you would want change it:
```
var options = {
   prePage: "Anterior", // Previous page button text
   nextPage: "Siguiente", // Next page button text
   firstPage: "Primero", // First page button text
   lastPage: "Ultimo" // Last page button text
}

[Primero] [Anterior] [ 7 ] ([ 8 ]) [ 9 ] [Siguiente] [ Ultimo ]
```

I changed too the file custom-pagination-table.js in order to exemplify these customizations' options.

### DIFF
--- a/examples/js/pagination/custom-pagination-table.js
+++ b/examples/js/pagination/custom-pagination-table.js
@@ -27,6 +27,10 @@ export default class CustomPaginationTable extends React.Component{
       sizePerPageList: [5,10], //you can change the dropdown list for size per page
       sizePerPage: 5,  //which size per page you want to locate as default
       paginationSize: 3,  //the pagination bar size.
+      prePage: "Prev", // Previous page button text
+      nextPage: "Next", // Next page button text
+      firstPage: "First", // First page button text
+      lastPage: "Last" // Last page button text
     }
 
     return (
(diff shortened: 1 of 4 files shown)


AllenFang:
@diazproduction, thanks your contribution, it's a nice patch :+1:  Anyway, I've released on v1.5.0

diazproduction:
Thanks! :D

Issue #260, opened 2016-02-09
On state change: Uncaught TypeError: Cannot read property 'props' of null


pirate:
Whenever I search my table (or update its state in any way), I get this error, and the table breaks.

`Uncaught TypeError: Cannot read property 'props' of null` on line `48` on on `TableColumn`'s `children`.

<img width="655" alt="screen shot 2016-02-09 at 2 29 15 pm" src="https://cloud.githubusercontent.com/assets/511499/12928100/a50bbac2-cf39-11e5-9512-08b6ef70d9f6.PNG">
Chrome dev console breakpoint set on line 48, showing the value of `children` immediately before the error.
<img width="668" alt="screen shot 2016-02-09 at 2 28 50 pm" src="https://cloud.githubusercontent.com/assets/511499/12928097/a0cd42fa-cf39-11e5-842b-d53c453336d8.PNG">

Fo some reason the if statement on line `47` seems to think `typeof "251" === "object"`, and it immediately errors on the next line when it tries to call `"251".props`, which of course doesn't exist since `"251"` is a string.  I've tried checking the type in console as well `typeof children >> "string"`, I'm stumped.  Any ideas?

----

**Code to reproduce:**
```javascript

export class CustomSortTable extends React.Component {
    render() {
        var {items, fields} = this.props;
        return <BootstrapTable data={items} search={true}>
            {fields.map((field, idx) => 
                <TableHeaderColumn
                    dataField={field.id}
                    dataSort={true}
                    key={field.id}
                    isKey={field.isKey || false}>
                    {field.name}
                </TableHeaderColumn>)}
        </BootstrapTable>
    }
};

var fields = [
    {id: 'id', name: 'ID', 'isKey': true},
    {id: 'my_value', name: 'My Value'},
]

var items = [{
     'id': '251',
     'my_value': true,
}, ...]

<CustomSortTable items={items} fields={fields}/>
```

AllenFang:
Thanks for bug report, I'll check this out after my vacation, need some time :)

pirate:
新年快乐！Enjoy your vacation  :), no rush.

AllenFang:
happy new year :)

AllenFang:
@pirate, which version of ```react-bootstrap-table``` did you used? and are you sure the code will produce the error? because I can't reproduce this followed by your example code

pirate:
[...]

PR #272, opened 2016-02-15
Fix null check bug

homuler:
This PR is related to #260 

When we update state, TableColumn's shouldComponentUpdate throws an error, so the table is not updated correctly.

```js
if(typeof children === 'object') {
   if(children.props.type === 'checkbox' || children.props.type === 'radio') {
```

`typeof` operator returns 'object' when it receives null value, so when we try to access to the props of `children`, a ReferenceError is thrown.

### DIFF
--- a/src/TableColumn.js
+++ b/src/TableColumn.js
@@ -20,7 +20,7 @@ class TableColumn extends React.Component{
       return shouldUpdated;
     }
 
-    if(typeof children === 'object') {
+    if(typeof children === 'object' && children !== null) {
       if(children.props.type === 'checkbox' || children.props.type === 'radio') {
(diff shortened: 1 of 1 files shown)


AllenFang:
@homuler, good, that's the evidence which I want to know. Thanks a lots :)

fastlorenzo:
Thanks for the fix :+1: 
Could you update npm so we have the fix available ? Thanks

AllenFang:
@fastlorenzo, I will release it in these day, sorry for a long wait :)

AllenFang:
published ```v1.5.1```. Thanks :)

fastlorenzo:
Thanks :)

Issue #273, opened 2016-02-15
Provide unminified production distribution


jochenberger:
I'd like to propose adding an unminified aggregated build (e.g. `dist/react-bootstrap-table.js`). This will make debugging much easier for users.

AllenFang:
HI @jochenberger, As I know, only one main entry point I can assign on npm package. Any idea?

jochenberger:
Sorry, I don't have much experience with node and npm. But, if I understand the question correctly, the minified version could still be the default and the unminified one could be an additional option. Or can you set it up so that the file is chosen depending on the `NODE_ENV` of the application?

AllenFang:
me too, i'm not sure about this issues, 

> Or can you set it up so that the file is chosen depending on the NODE_ENV of the application?

I think it's impossible, i'll spend some time to find the solution, but the priority is lower, sorry

AllenFang:
if you find any case like you said, please let me know or PR is welcome :)

jochenberger:
Closed in favor or #277

dana2208:
Hi,

While running the project (either with gulp prod or gulp example-server) in chrome you can see in the developer tool in the source tab the folders "webpack://" > "." > "src". In the "src" you can see all the sources and debug there.
Isn't it what you wanted?
![image](https://cloud.githubusercontent.com/assets/10193337/13081556/ed2382b2-d4d5-11e5-994d-44783821b449.png)

jochenberger:
@dana2208 no, I want to be able to use my Java web framework's minification/aggregation capabilities.

AllenFang:
@jochenberger, you used Java web framework to minify javascript code? what tool you used?

jochenberger:
I'm using [Apache Tapestry](http://tapestry.apache.org/) together with https://github.com/eddyson-de/tapestry-react. The minification is done via [Google Closure Compiler](https://developers.google.com/closure/compiler/).

Issue #217, opened 2016-01-19
Debug the original React files


dana2208:
Hi,

Is there a way I can debug the original react files when developing my application that uses your grid?

Thansk

AllenFang:
@dana2208, ok, I'll add this feature, but  I'm busy in recent, so need sometime to do it. If any PR is welcome :)

AllenFang:
@dana2208, use webpack ```dev-tool: 'eval-source-map'``` to provide the source map, please check this out on newest version

dana2208:
hello,
Thanks a lot for this! I am finally able to debug with the web developer tool.
However there is a weird behavior:
In the web developer tooI I set a breackpoint in the generated BootstrapTable.js but it doesn't get hit until I also set a breackpoint in my own jsx file, the one that renders BootstrapTable.
According to this [stackoverflow answer](http://stackoverflow.com/a/34247061/2334891), maybe you should add:
```
debug: true,
devtool: "#eval-source-map",
```
I guess you don't want to put this in your webpack for production so maybe you could write another webpack for development use only.

Thanks

AllenFang:
> so maybe you could write another webpack for development use only

Just a little question,
you mean I can add ```debug:  true``` in my dev webpack for only development?
Thanks :)

jochenberger:
If I understand https://webpack.github.io/docs/configuration.html#devtool correctly, the `*eval*` devtools should not be used for production. In any case, they cause issues if the resulting files are passed through a minifier. For example, I get: "RequireJS error: define: exports is not defined, modules react-bootstrap-table"

PR #277, opened 2016-02-16
Add unminified build

jochenberger:
This adds another webpack configuration for an unminified production build. This PR supersedes #273.

### DIFF
--- a/webpack.production.min.config.js
+++ b/webpack.production.min.config.js
@@ -1,13 +1,13 @@
 var webpack = require('webpack');
 
 module.exports = {
-  devtool: 'inline-source-map',
+  devtool: '#eval-source-map',
   entry: {
     'react-bootstrap-table': './src/index.js'
   },
   output: {
     path: './dist',
-    filename: '[name].js',
+    filename: '[name].min.js',
     library: 'ReactBootstrapTable',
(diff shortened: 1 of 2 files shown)


webpack.production.min.config.js
```diff
@@ -0,0 +1,43 @@
+var webpack = require('webpack');
+
+module.exports = {
+  devtool: '#eval-source-map',
```
jochenberger: This still uses `eval`, which should probably be avoided in production (see #217), maybe this shhould go into another build, maybe `.debug.js`.

AllenFang:
@jochenberger, I think this PR will generate two result file in ```dist``` folder, one in minified and the other is non. but for the user who install by npm, they always got the minified one. Because  the following line defined in ```package.json```:
```"main": "./dist/react-bootstrap-table.min.js",```

Which means you used ```react-bootstrap-table``` by imported scripts to HTML instead of npm, am I right?

jochenberger:
Yes, almost, I use it as an AMD module via Require.js. I can control which file to use in my app's configuration.

AllenFang:
@jochenberger, thanks your contribution, this PR is merged in v1.5.2

Issue #291, opened 2016-02-22
Column-filter failed when using default value and pagination


AllenFang:
Following is error message: 
```
Uncaught TypeError: Cannot read property 'getSizePerPage' of undefined
```
[There](https://github.com/AllenFang/react-bootstrap-table/pull/270#issuecomment-187059679) is the discussion

AllenFang:
Fixed on v1.5.3

PR #292, opened 2016-02-22
Bug happens when using defaultValue of a filter together with pagination

dana2208:
For a better fix, need to handle pagination states in Bootstrap table
Bug reference is #291


AllenFang:
@dana2208, I'll close this PR due to I almost move pagination state to BootstrapTable. btw, I found the origin column filter broke after v1.5.1, because the filter implementation changed. You can check this [example](https://github.com/AllenFang/react-bootstrap-table/blob/master/examples/js/manipulation/filter-table.js)

PR #301, opened 2016-02-25
Added mouse enter and leave events for table and rows

MichelCarroll:
I added event hooks for mouse events on rows and on the table itself.


AllenFang:
@MichelCarroll, LGTM, I'll test this in these day. Thanks your contribution

AllenFang:
Released on v1.5.4. Thanks your contribution

PR #303, opened 2016-02-26
Pagination formatting fix

k-nilsson:
… provided.

When no sizePerPageList prop is supplied, or if there's only one element, the class on the pageList is now properly set to reduce instances of button overflow.


AllenFang:
@k-nilsson, LGTM. Thanks your contribution :+1:

AllenFang:
Released on v1.5.4 :)

PR #306, opened 2016-03-01
Clear timeout on number and text filters when component unmount

dana2208:


AllenFang:
@dana2208, thanks you :)

PR #311, opened 2016-03-03
Search improvs

staminaloops:
- **new prop on TableHeaderColumn:** `searchable`;
 EDIT: default is true.
- **improv search performance**:  
forEach is not performant and for...in is even worse. For loops are ugly, but sometimes it's the better choice. http://jsperf.com/for-vs-foreach/66
- **search by number** to exact match that number - I suppose that if we search for a number, we would want that exact number - if you dont agree, we can take this one.

:)


staminaloops:
Maybe if we put something like 

`if (!hidden || searchable)`

it wouldnt break people's code again, and we could make a warning like 'on the next major version you have to explicitly tell which columns are searchable'

AllenFang:
@staminaloops, nice patch!! Just a little question,  I think it's better if the default value of ```searchable``` is true(for users)? or did you have any consideration on why default is false?

staminaloops:
Yes, maybe if the searchable default is true, the users just have to defined which columns they dont want to search, it makes sense. In these case, do you want the hidden fields to be searchable by default?

```
if (!hidden && searchable)
```

or just 

```
if (searchable)
```

AllenFang:
@staminaloops, make the hidden column can't be searchable by default, thanks :+1: I'll merge this in these day :ship:

AllenFang:
@staminaloops, I think make hidden column cant be search by default is good.

AllenFang:
@staminaloops, another question when I taking a code review, why add a type [checking](https://github.com/staminaloops/react-bootstrap-table/commit/d87f33911a2b88ebb34150e0ab5274c32f378c05#diff-6dae34a6b5be3960c40ef0a8e0396673R370) for number in search?

staminaloops:
The type check is part of a new feature.
If we have an array of objects like in the examples:
```js
    {
      id: id,
      name: "Fruit name is " + fruits[i],
      price: 2100 + i
    }
```
The price is a number, and people can search for that exact number, instead of search for every columns that includes part of that number - I think that's what people want most of the time, but as I said, we can take this feature away.
image

Regarding hidden columns being not searchable by default - I've being thinking, and it doesnt make sense. There is only one reason to include hidden columns - its to make them searchable. If you dont want to show that columns, you just dont add them to Bootstrap table.
So I think that hidden columns should be searchable by default, like all the other columns.
This makes sense to you?

AllenFang:
@staminaloops, I thinks we should not need the type checking on search, because, the search inspired by [DataTables](https://www.datatables.net/), if user want a search for type, they can use column filter. And I will patch your PR for this feature away

btw, I agree your default searchable of hidden column strategy. Thanks !

AllenFang:
I've done this PR, check the newest version and thanks your contribution :+1:

AllenFang:
> Regarding hidden columns being not searchable by default - I've being thinking, and it doesnt make sense. There is only one reason to include hidden columns - its to make them searchable. If you dont want to show that columns, you just dont add them to Bootstrap table.
So I think that hidden columns should be searchable by default, like all the other columns.


@staminaloops, I forgot I need make search on hidden column configurable, so I'll make 
```
if (!hidden && searchable)
```
to 
```
if (searchable)
```
Means all column(hidden or display) will be searched, and just turn off by ```searchable={false}```
Although your explanation is make sense, there always some requirement that user need. Feel free to discuss with me :)

staminaloops:
Yes @AllenFang, that was what I was talking about and what I intended to do

```
if(searchable)
```

I thought you had done it before you merge it :smile:

AllenFang:
haha, I think I need to improve my english skill. Anyway, Thanks your help! You did a great job

PR #312, opened 2016-03-04
Reusable sortFunc

r1g0:
I added sortField to sortFunc call to have the field to which the sort is being applied.

I needed this to reuse the number sort and case insensitive sort on multiple columns and different tables.  I also added a demo of a use case.


AllenFang:
@r1g0, thanks your contribution, I'll check this out in these day :+1:  :+1:

AllenFang:
Released on newest version, Thanks again :)

r1g0:
Great!  Thanks for this great lib.

PR #328, opened 2016-03-11
Set the column text as the title of the column header

jfremy:
With small columns, it's not sure the full header text can fit.
To account for this case, set the title attribute of the column header to the text of the column header.
That way, if the text is not completely visible, hovering the mouse on the header will display the text.

This makes the assumption that this.props.children is a string, but that seems to be already the case for the filters seem to expect that.

Otherwise, it might be convenient to allow manually setting the title attribute as a property of TableHeaderColumn


AllenFang:
LGTM

AllenFang:
Released on ```v2.0.1```, Thanks :+1:

PR #333, opened 2016-03-13
-Update ignores to remove generated files. These files will still be …

narfman0:
…pushed to npm with the added .npmignore file. This should reduce conflicts for merges and diff clutter.


AllenFang:
@narfman0 because user could download from tag or zip for using ```react-bootstrap-table```. so I think I will keep dist/ folder available on git

AllenFang:
I've merged this PR and will released on next version. Thanks :)

narfman0:
Ok I'll avoid gulp prod in the future for commits, thanks!

There should be a 'releases' feature in github documented here: https://help.github.com/articles/creating-releases/ for if a user could download directly.

I'll close this for now though since you don't want it, thanks for your input!

AllenFang:
@narfman0, no..  I've did some change and merged it in my local.

AllenFang:
btw, thanks your github documents and I'll study hardly for a correctly released action. Thanks :+1:

narfman0:
Oh ok, great! Thanks! :)

PR #335, opened 2016-03-14
example on how to use floats in grid

DexterHaslem:
it was initially unclear to me how to get floats to behave properly with sort + filter, so hopefully this helps someone


AllenFang:
@DexterHaslem, a good example! Thanks your contribution:)

PR #339, opened 2016-03-15
workaround for filters toggling sort

### LINKED ISSUE #336: interacting with filters toggles column sort
If you have a column header setup with dataSort and a filter, messing with the filter continually toggles sort. for example:

    <TableHeaderColumn dataField="num" dataSort={true} filter={{ type: 'NumberFilter'}}>

I tried to fix by putting a check [here](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/TableHeaderColumn.js#L19)

I tried checking if the target was this.refs["header-col"], however that is too restrictive, and clicking on span with column label would not sort. Any ideas on how to make filter + sort more usable?

DexterHaslem:
fixes #336

### DIFF
--- a/src/TableHeaderColumn.js
+++ b/src/TableHeaderColumn.js
@@ -16,6 +16,7 @@ class TableHeaderColumn extends React.Component{
 
   handleColumnClick(e){
     if(!this.props.dataSort)return;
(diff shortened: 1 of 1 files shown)


AllenFang:
@DexterHaslem, Fixed on v2.0.2 with a better solution

DexterHaslem:
Thanks!!

PR #341, opened 2016-03-15
Fix 'document is not defined' when used with server side rendering

staminaloops:
I was getting this error with SSR.
I added a function with a type check in utils.js if you need to use it in other places in the future.


AllenFang:
@staminaloops, merged and released on ```v2.0.2```. Thanks your contributions:)

PR #352, opened 2016-03-23
improve the example about selection externally last page event issue

madeinfree:
improve the example about selection externally-managed-selection last page event issue, because the last page always back to the first page.

1. added the currPage state for externally-managed-selection to avoid if the state was changed, and get the correct page.

2. added the handleSelectRow result call back, currPage parameter.

I think changed this issue maybe improve the user experience.

### DIFF
--- a/examples/js/selection/externally-managed-selection.js
+++ b/examples/js/selection/externally-managed-selection.js
@@ -23,14 +23,21 @@ export default class ExternallyManagedSelection extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      selected: []
+      selected: [],
+      currPage: 1
     };
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


AllenFang:
@madeinfree, I know this issue, but just dont fix it.  I think you work hardly on these examples. However, thanks your contribution :)

madeinfree:
Ok, thanks @AllenFang

madeinfree:
@AllenFang I try to revised this PR, and add the ref for BootstrapTable it looks like work, is that suitable ?

AllenFang:
@madeinfree, I thinks it's better than before, I'll merge it today. Thanks :+1:

madeinfree:
@AllenFang Thanks :+1:

Issue #353, opened 2016-03-23
Selected row page


dana2208:
Hello @AllenFang ,

I have selected a row from an external action by passing the selected array to the `selectedRow` props of the `BootstrapTable`.
Is there a way I can know in which page is the selected row?
I would like to be able to jump to that page when it is selected, either right after the selection or at the same time as it is selected by passing the `page` in the `options` props of the `BootstrapTable` together with  the selected array.

Thanks

AllenFang:
HI @dana2208, The question is you want to get these pages which contain selected rows, right?
I think there's no way to do it. How about I expose an API on ```BootstrapTable``` which can let you get these selected row pages?

AllenFang:
btw, did you need to know the page which every selected row located?

dana2208:
Hi,

I would like either to get the page numbers
1. before passing the array of selected row keys,
2. or after passing the array of selected row keys.
First solution would be better because I could update props of the `BootstrapTable` (`selected` and `page`) in the same `setState` call but second solution will be OK too.

Currently I am working in mode "radio" so I need only one page number, but it is better to support already the "checkbox" mode (multi-selection) so if you implement it, it is better to get an array of the pages.

What do you think?

madeinfree:
Hi @dana2208, I guess you want to click the checkbox and get the current page of selected row key immediately ?

dana2208:
Hi, 

No, my selection is done externally. 
I am clicking an entity on a map and i want it selected in the bootstrap table grid that is displayed below the map.
So when clicking on the map i get the id of the entity which i put in the selected array of the selectRow prop of BootstrapTable. I would like to get the page in which this selected row key is in the grid so that I also can go to this page through the page props.

AllenFang:
@dana2208, I understand, maybe like I said, just give you a exposed API for you to get the page of selected row key, I think it should solve this issues. if ok, I'll done it in these day :)

AllenFang:
hi @dana2208, @madeinfree is a new contributor who I found for helping me. Because, I'm busy in these day(always busy...lol).  btw, I'll ask him support this feature at next version. Feel free to discuss with him and me :+1:

madeinfree:
@dana2208 Ok, I understand and I will support that feature.

@AllenFang Thanks, i will try it.

dana2208:
[...]

PR #358, opened 2016-03-24
Page by rowkey

madeinfree:
Hi @AllenFang, thanks for your guided !!
This is about #353 I added the example page call expose-api-table, but the name and file path made me confusing.. I don't know that suitable ?


AllenFang:
Good patch

AllenFang:
Released on ```v2.1.0```. Thanks :)

Issue #349, opened 2016-03-22
Unresponsive page when searching large tables


JeremyFossTemple:
If the table is large then the page becomes unresponsive during search.  Maybe consider offloading search to a webworker, or searching for a set amount of time, then window.setTimeout(continueSearch)

AllenFang:
@JeremyFossTemple, yes, its a performance issue on table search, #330 have asked a delay when user  want to search, I'll done this tuning in next version. About the performance issues, I'll try to solve in near future. Thanks :)

AllenFang:
@JeremyFossTemple, we've support a delay on search, I thinks it a workaround for this issue. Check [example](https://github.com/AllenFang/react-bootstrap-table/blob/master/examples/js/manipulation/debounce-search-table.js)

Issue #330, opened 2016-03-13
delay for the search feature


dana2208:
Hi,

I think there should be a "searchDelay" property (500ms for default) for the search feature so that not each new "key down" triggers a search.
This will improve the search feature performance.
Of course this delay could be overwritten.

What do you think?

AllenFang:
@dana2208, agree, it's a good suggestion, but I think the default is zero is better for the user who has been use search functionality.

AllenFang:
@dana2208, we've support a delay on search, I thinks it a workaround for this issue. Check [example](https://github.com/AllenFang/react-bootstrap-table/blob/master/examples/js/manipulation/debounce-search-table.js)

PR #360, opened 2016-03-25
Search delay feature

madeinfree:
Hi, @AllenFang this is my solution,
The search delay time feature solution for #330 #349.
use the `options` set `searchDelayTime`(type integer) in it, and it can helpful for set search delay.

it work well with my example.


AllenFang:
@madeinfree, Thanks, I've already merged it :)

AllenFang:
Released on ```v2.1.0```. Thanks :)

PR #367, opened 2016-03-28
Cells with no dates will not cause any error anymore and will be filt…

dana2208:
…ered if a date filter is used.
Only TableDataStore.js was updated.


AllenFang:
Thanks @dana2208

AllenFang:
Release on ```v2.1.1``` :)

PR #368, opened 2016-03-28
When using a custom sort function, pass any other extra data.

dana2208:
The example shows how a column that displays enum values can be sorted according to those values and not the enum number.
const qualityType = {
  0: 'good',
  1: 'bad',
  2: 'unknown'
};
Indeed, instead of sorting item tables according to the enum keys 0, 1 and 2 it is possible to sort now by the enum values "good", "bad" and "unknown" alphabetically which makes more sense since this is what the user can see.
ASC order will show "bad" items, then "good" items then "unknown" items (alphabetical order is respected), instead of showing "good", then "bad", then "unknown".


AllenFang:
Thanks @dana2208 :)  It's a good example, I'll add it to the docs in website also.

PR #384, opened 2016-04-04
use custom render method for the caret

vladelpmk:


AllenFang:
@vladelpmk, good patch.  Thanks your contribution :)

AllenFang:
Released on ```v2.1.2```

PR #385, opened 2016-04-04
Don't change sort on props update

just-paja:
We have a react-redux application where component above the table re-renders after click on a table row. The table receives same props, but sort is reset to default. I belive it is either a bug or bad name for `sortName` and `sortOrder` options - the default shouldn't be prefered over user choice.

Documentation says:
sortName : String : Assign a default sort field.
sortOrder : String(asc/desc) : Assign a default sort ordering.

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -146,8 +146,8 @@ class BootstrapTable extends Component {
       page = 1;
     }
     const sortInfo = this.store.getSortInfo();
-    const sortField = options.sortName || (sortInfo ? sortInfo.sortField : undefined);
-    const sortOrder = options.sortOrder || (sortInfo ? sortInfo.order : undefined);
+    const sortField = sortInfo ? sortInfo.sortField : options.sortName;
+    const sortOrder = sortInfo ? sortInfo.order : options.sortOrder;
     if (sortField && sortOrder) this.store.sort(sortOrder, sortField);
     const data = this.store.page(page, sizePerPage).get();
     this.setState({


AllenFang:
@just-paja, sorry I'm not familiar with redux, but could you please show me your code snippet?

I thinks this issues should be a bad description for ```sortName``` and ```sortOrder```, and I guess you always render table with option with the same ```sortName``` and ```sortOrder```.

At first, this both sort option if absolutely for setting default sort, but now it seems to be a sort state controlled by user(**Let you manage the sort state on your component, not table itself**), This is what you encountered and reason for why I put ```options.sortName``` and ```options.sortOrder``` at first statement instead of second.

So you can check this [example](https://github.com/AllenFang/react-bootstrap-table/blob/master/examples/js/sort/sort-hook-table.js), I thinks this can solve your issue.

Anyway, feel free to discuss with me :+1:  
BTW, If you agree it, I'll improve the documents.

just-paja:
@AllenFang, thanks for your answer. 

It is my understanding that `sortName` and `sortOrder` are interactive props, so it might be nicer to go the same way React inputs, where you have `value` for controlled elements and `defaultValue` for interactive components.

I tried [implementing your example](https://gist.github.com/just-paja/c2797dd43eee9e7d9e22870c8e135099) as a wrapper, but it does extra rerendering. 

1. User changes sort value, table redraws its contents
2. Wrapper receives callback, sets new state and renders the table again

If it was done similar to `defaultValue` behaviour, table would not rerender it's contents.

So what do you think about this? We change documentation for `sortName` and `sortOrder` to way they are and add `defaultSortName` and `defaultSortOrder` to

AllenFang:
> So what do you think about this? We change documentation for sortName and sortOrder to way they are and add defaultSortName and defaultSortOrder to

Exactly what I think. But it's a **api change**, we need handle it carefully.  I'll think some solution in these day, may not released in near future.

BTW, the [React inputs](https://facebook.github.io/react/docs/forms.html) is a good idea

PR #388, opened 2016-04-05
Bug/filter ignores sort

just-paja:
User changes sort, table gets sorted. He then selects filter value, and sort gets ignored. I've put together `select-filter-with-sort` example where you can see it.

### DIFF
--- a/examples/js/column-filter/demo.js
+++ b/examples/js/column-filter/demo.js
@@ -5,6 +5,7 @@ import TextFilterWithDefaultValue from './text-filter-with-default-value';
 import RegexFilter from './regex-filter';
 import SelectFilter from './select-filter';
 import SelectFilterWithDefaultValue from './select-filter-with-default-value';
+import SelectFilterWithSort from './select-filter-with-sort';
 import NumberFilter from './number-filter';
 import NumberFilterWithDefaultValue from './number-filter-with-default-value';
 import NumberOptionsFilter from './number-options-filter';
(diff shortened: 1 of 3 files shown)


AllenFang:
@just-paja, thanks your contribution, I thinks it's a big bug since we supported filter on header and your solution is not bad. Actually, the root cause is on [here](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/store/TableDataStore.js#L206).  We always use ```this.data``` to filter, you can see the other function in ```TableDataStore```(ex: remove, search, add  etc.), The correct way is like following:
```js
const currentDisplayData = this.getCurrentDisplayData();
this.filteredData = currentDisplayData.filter( row => {
   //....
}
```

Anyway, You did a great job and do you want to keep to fix it? I can let you fix this issue follow above-mentioned :)

just-paja:
@AllenFang, thanks, but your suggestion actually does not work. You would filter filtered data and eventually get to empty array. I think that my way is efficient. It sorts only when requested and right after user filtered data.

AllenFang:
Sorry, I got a misunderstand on filter mechanism. I think your solution is well!

thomasturrell:
Hey @AllenFang this would be really useful.  I've encountered this issue too. (full disclosure, I work with @just-paja)

Keep up the good work, this is a really useful project.

AllenFang:
Release on ```v2.1.2``` :)

Issue #392, opened 2016-04-06
Event Object passed to onRowSelect callback


humblepig:
Would it be possible to pass the event object as the third parameter to the onRowSelect callback?  This would allow user to check to see if there was a key pressed when the row was clicked.  There may be other use cases for this.

AllenFang:
@humblepig, I'll add it

madeinfree:
I try to add the event for OnRowSelect, see the #399

AllenFang:
@humblepig, fixed on ```v2.1.3```, the event object will be passed as third argument.

PR #399, opened 2016-04-11
OnRowSelect callback

madeinfree:
for #392 solution.


AllenFang:
@madeinfree, thanks your contribution. But I've solve this issue in my repo and I'll release it today. BTW, #392 want to pass event object on ```onRowSelect``` instead of ```onRowClick```, they are different.  And according [this](https://github.com/AllenFang/react-bootstrap-table/issues/392#issuecomment-208259261), in best practice, we should put event object as first argument like you said, but if we change it like that will cause all the user need to change too. Anyway, we can think about it again if we bump major version to higher in future.

madeinfree:
@AllenFang Thanks, I attention it next time  👍

Issue #398, opened 2016-04-11
Editable property input box can't copy value


rbalicante882012:
When I click the cell the editable input box did not copy the value. But, when I used textarea, it works.

madeinfree:
could you more detail description your problem or have some example for me to reproduce it ?
it work well in my example.

rbalicante882012:
Hello @madeinfree. 
I add validator property for editable and I made handler but I won't copy the value of input box only. Only if I put type property.

AllenFang:
@rbalicante882012, it's a bug, I'll fix it quickly

rbalicante882012:
@AllenFang thanks :+1:

madeinfree:
Hi, @rbalicante882012 @AllenFang I push the PR above #401, you can review it.

rbalicante882012:
Thanks @madeinfree  :+1:

rbalicante882012:
@madeinfree can I ask? 

What time @AllenFang will check it @madeinfree ?

AllenFang:
@rbalicante882012, I'll do it in these day, before thursday ;)

madeinfree:
@rbalicante882012 sure, feel free to discuss

rbalicante882012:
Hi @madeinfree can I manipulate checkbox in the table header ? I want it checked = false.

AllenFang:
you want to set the checkbox on header programmatically?

rbalicante882012:
Yes @AllenFang.

AllenFang:
@rbalicante882012, what version did you used?  I try to call ```cleanSelected``` and it works(all checked box will uncheck, include the header one).  BTW, could you show me your code?

Thanks :)

rbalicante882012:
@AllenFang I want header to checked to false. Can I manipulate it?

AllenFang:
No, the header checkbox depend on user operation.

rbalicante882012:
@AllenFang I want it to clear after I click the delete button.

AllenFang:
let me guess your scenario, 
1. user select some column, so header column checked
2. user click delete button to drop selected row
3. header column checkbox remain checked?

some question I want to ask

1. the delete button is means you enable ```deleteRow```? or you implement a delete button by yourself and also delete row by yourself?
2. I've test and it work when I enable ```deleteRow```.

rbalicante882012:
@AllenFang, 
I created my own delete function and I want to uncheck the header check box. :)

[...]

PR #401, opened 2016-04-12
Fixed validator type

madeinfree:
Hi, This is the fixed for #398 

BTW this time I added a little features for table row validator notifier, if you want to use the notifier for table text field, but must be use object like:
```javascript
function jobNameValidator(value) {
  if (!value) {
    return {
      validate: false,
      tip: 'Job Name is required!'
    };
  } else if (value.length < 10) {
    return {
      validate: false,
      tip: 'Job Name length must great 10 char'
    };
  }
  return true;
}
```
and the notifier work on table row.
or see the example /examples/js/advance/validator-row-notifier-table.js

I didn't fine other bug for it now, but maybe it will affect other features..


AllenFang:
I'll review it, Thanks :)

Issue #397, opened 2016-04-09
React 15 Support


staskorz:

madeinfree:
Hi @staskorz,

I think it not easily to change the version now, because some of dependencies have to upgrade or change it ... and React changed the version 0.14.x jump to 15.0.0 this time, so we must be solve the dependencies issue now ~

I think it could support React 15 quickly !!

cc @AllenFang

staskorz:
Hi @madeinfree,

Thanks for the "quickly" comment, I've change the title :-)

Re the dependencies: this package only depends on `classnames` and `react-toastr`. The later depends on React 0.14, I've just opened an issue in its repository (tomchentw/react-toastr#47).

Technically it's possible to depend on multiple `react `and `react-dom` versions like so:

`"react": "^0.14.3 || ^15.0.0"`

As till now both packages were running without React warnings, they should keep working with React 15, at least according to the Release Notes.

AllenFang:
Released on ```v2.0.0```

PR #407, opened 2016-04-14
Fix filter default value

jfremy:
When the component checks whether there is a default filter applied, it fetches the value from defaultValue.
This does not work. defaultValue on the input component is empty.
The value property however, contains the correct value.

Without this fix, the defaultValue property for the filter is not applied for text, regex and dates

### DIFF
--- a/src/filters/Date.js
+++ b/src/filters/Date.js
@@ -62,7 +62,7 @@ class DateFilter extends Component {
 
   componentDidMount() {
     const comparator = this.refs.dateFilterComparator.value;
-    const dateValue = this.refs.inputDate.defaultValue;
+    const dateValue = this.refs.inputDate.value;
     if (comparator && dateValue) {
       this.props.filterHandler({ date: new Date(dateValue), comparator }, Const.FILTER_TYPE.DATE);
     }
--- a/src/filters/Regex.js
+++ b/src/filters/Regex.js
@@ -19,7 +19,7 @@ class RegexFilter extends Component {
   }
 
   componentDidMount() {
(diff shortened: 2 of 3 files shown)


AllenFang:
HI @jfremy, I've test the original ```defaultValue``` and it work for me. Could you please try this [example](https://github.com/AllenFang/react-bootstrap-table/blob/master/examples/js/column-filter/text-filter-with-default-value.js)?

I think there's something missing, and did you have an example which can reproduce this issue?  And what browser you used?  

Thanks :)

jfremy:
This seems to be linked to the "environment". I use react 15.0.1, react-router 2.2.4, react-bootstrap 0.28.5, react-bootstrap-table 2.1.4 and if I take the example component from text-filter-with-default-value.js and put it into my app, then it does not work while it works fine in your example application.
I'll continue investigating what is the trigger

jfremy:
It seems to be React 15. If I run your example with React 15 & react-dom 15, then the example fails the same way it does fail in my app.

jfremy:
It looks like that React 15 does not set an initial value attribute on the input element (probably a bug) even if defaultValue is set.
So defaultValue is always the empty string.

AllenFang:
@jfremy, sorry for a lately reply and thanks your greatly investigation. I think I'll make ```react-bootstrap-table``` compatible to React@0.15 in near future(#397) and then I'll try to fix this issue. 

Thanks :+1:

AllenFang:
@jfremy, this is PR has been released on ```v2.2.0```. Thanks your contribution :)

PR #417, opened 2016-04-23
Show position + total lines displayed

jfremy:
Hi, this is a proposal for a small improvement to the pagination feature
I've added an option (paginationShowsTotal) that if set to true will display a short text giving the current lines displayed + the total number of elements (for instance: showing rows 5 to 9 of 45)

This is convenient when using the pagination (since only react bootstrap table knows what is the current page) but also when used in conjunction with filtering (it tells you how many lines match the filter which again we can only access from within react bootstrap table)

Let me know what you think

### DIFF
--- a/examples/js/complex/app.js
+++ b/examples/js/complex/app.js
@@ -72,6 +72,7 @@ const cellEditProp = {
 };
 
 const options = {
+  paginationShowsTotal: true,
   sortName: 'name',  // default sort column name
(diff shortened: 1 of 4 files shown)


AllenFang:
@jfremy, that's a good feature for pagination, I'll merge it in next version 👍 

Thanks :)

AllenFang:
@jfremy, Release on ```v2.1.5```.  Thank a lots

PR #418, opened 2016-04-26
custom data accessor

### LINKED ISSUE #50: dataField can't access subobject
I have an object which is like that

```
{
  "person_id": 1,
    "username": "UserTest",
    "created_at": "2015-09-10T21:13:56.000Z",
    "updated_at": "2015-09-10T21:13:56.000Z",
    "deleted_at": null,
    "Person": {
      "id": 1,
          "firstname": "UserTest",
          "lastname": "Obama",
          "note": "DevOps",
          "email": "iam@obama.com",
          "created_at": "2015-09-10T21:13:56.000Z",
          "updated_at": "2015-09-10T21:13:56.000Z",
          "deleted_at": null
    }
}
```

I can't access the subojbect using

```
<TableHeaderColumn dataField="Person.firstname">Firstname</TableHeaderColumn>
```

Is there any workaround not specified in the doc ?

jdeniau:
I created a custom data accessor. The purpose is the same as #50 but it provide a more flexible way of doing it.

You only need to pass a `dataAccess` parameter to the `TableHeaderColumn`. The function takes the data as parameter and you can return whatever you want.

This way, it fixes nested properties, but also if you use something like immutablejs it works fine.

I kept the `dataField` attribute as I saw you use it for a lot of purpose, but as it is not used for accessing data, it's more like an id I think.

I make it work with simple filter, it does not work with `search` because you iterate over each object instead of each column here.

Fixes #50


jdeniau:
The tests are failing on an npm error I think between react-toastr which now requires react >= 15 and your component and / or react-dom

AllenFang:
@jdeniau, thanks your contribution :)  I'll check this out in these day

welcoMattic:
Is someone still working on this PR? Fixing conflicts?

jdeniau:
I'm not, I do not use this package anymore but feel free to update it

welcoMattic:
@AllenFang I think you can closes this PR since this one (https://github.com/AllenFang/react-bootstrap-table/pull/508) is open.

jdeniau:
Closing in favor of #508

Issue #419, opened 2016-04-26
MAJOR ISSUE: Need to used fixed version in package.json


invalidred:
Hi Allen,

In your package.json you are using ^ (Compatible with version) everywhere. For example
    "react": "^0.14.3",
    "react-bootstrap": "^0.27.2",
    "react-dom": "^0.14.3",

We are facing an issue where npm installs version 15.0.1 on our server under react-bootstrap-table folder. Since our project uses react 0.14.8, we are getting the following issue

https://gist.github.com/jimfb/4faa6cbfb1ef476bd105.

Is it possible to update package.json to used fixed versioning. So it will become:
    "react": "0.14.3",
    "react-bootstrap": "0.27.2",
    "react-dom": "0.14.3",

This way there wont be incompatible react versions installed?

bindiababy:
we are facing similar issue - our app uses react 0.14.8 , but react-toastr@2.6.0 which is brought in by react-bootstrap-table expects react 0.15 and this  causes our app to fail with error - you have multiple copies of React loaded (details: https://fb.me/react-refs-must-have-owner). Is it possible to have react-bootstrap-table backward compatible with 14 version?

AllenFang:
hi dears, it's duplicate with #397, I'll make ```react-bootstrap-table``` compatible with react@0.15 and react@0.14.

AllenFang:
FIx on ```v2.2.0```. Thanks :0

PR #423, opened 2016-04-27
React 15 support

### LINKED ISSUE #397: React 15 Support

staskorz:
Adds React 15 as a dependency, so now it's possible to either depend on React 0.14 or 15.0. Closes #397 and #419


AllenFang:
@staskorz, thanks your contribution, I'm doing this patch, because there's some bug after upgrading. But just remain this PR, I'll merge it.

staskorz:
@AllenFang, thanks for the quick reply.

AllenFang:
Released on ```v2.0.0```

Issue #422, opened 2016-04-27
property title for each column


sarzijan:
Hi @AllenFang ,

Is it possible to add new props e.g. "columnTitle" for component TableHeaderColumn, which will set html property title for each column?

I am using your awesome component and if I use it without dataFormat, the css style with ellipses are applied, then I want to see the whole text in title when I hover over the column.

Yes, I can use dataFormat and wrap it to the span with the title property, but then I have to call the formatter for each column, where the title should be sown. and also the default css style is not used and I have to create my own for that...

Thanks a lot

AllenFang:
@sarzijan, I can support ```columnTitle``` for this requirement :)

AllenFang:
@sarzijan, thanks your contribution, this patch has been released on ```v2.3.1```. Thanks 👍

PR #436, opened 2016-05-04
column tilte property for individual columns issue #422

sarzijan:
Added bool property columnTitle , issue #422, if true the same value used in html property title


AllenFang:
thanks @sarzijan, I'll check this out :)

PR #445, opened 2016-05-10
Fix sorting for null values

Kurtas:
Columns with some null values are not sorted correctly.


AllenFang:
thanks, @Kurtas, I'll check this out :)

AllenFang:
Released on ```v2.3.2```. Thanks @Kurtas

PR #447, opened 2016-05-11
Add support for pagination start index

ianks:
I am using a service which uses 0-based indexing for paginating their API. I wanted a way in which I would simply map the state of my URL to params for their API. Rather than manually decrementing the `page` param each time I wanted to make an API, I think it would be cleaner, more robust, and simpler to maintain to simply allow for non-zero based indexing in react-bootstrap-table.

What this PR does is allow for the configuration option `pageStartIndex`. which allows the user to specify which page to begin paginating from.

If there is anything that needs to be changed, please let me know!

Cheers,
Ian

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -142,7 +142,13 @@ class BootstrapTable extends Component {
     const { options, selectRow } = nextProps;
 
     this.store.setData(nextProps.data.slice());
-    let page = options.page || this.state.currPage;
+
+    let page;
+    if (options.page !== 'undefined') {
+      page = options.page;
+    } else {
+      page = this.state.page;
+    }
 
     if (this.isRemoteDataSource()) {
       this.setState({
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -144,7 +144,7 @@ class BootstrapTable extends Component {
     this.store.setData(nextProps.data.slice());
 
     let page;
-    if (options.page !== 'undefined') {
+    if (options.page != null) {
       page = options.page;
     } else {
       page = this.state.currPage;
@@ -399,7 +399,7 @@ class BootstrapTable extends Component {
         isSelected ? this.store.get() : []);
     }
 
-    if (typeof result === 'undefined' || result !== false) {
+    if (typeof result == 'undefined' || result !== false) {
       if (isSelected) {
         selectedRowKeys = this.store.getAllRowkey();
       }
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -39,7 +39,9 @@ class BootstrapTable extends Component {
 
     this.state = {
       data: this.getTableData(),
-      currPage: this.props.options.page || 1,
+      currPage: this.props.options.page ||
+                this.props.options.pageStartIndex ||
+                Const.PAGE_START_INDEX,
       sizePerPage: this.props.options.sizePerPage || Const.SIZE_PER_PAGE_LIST[0],
       selectedRowKeys: this.store.getSelectedRowKeys()
     };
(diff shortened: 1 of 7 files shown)


AllenFang:
HI Ian, thanks your contribution :+1:   I'll check this out in these day. it's nice feature for ```react-bootstrap-table``` :)

ianks:
There is currently a bug with remote-paging. For some reason, the currPage is not being set correctly in BootstrapTable. This can be re-created by testing out the [Remote Paging Example](http://localhost:3004/#/examples/remote).

I've been banging my head against the wall about this one, because setState is called with the correct values here...

https://github.com/ianks/react-bootstrap-table/blob/8acaf42c46fd949adb7b455508e11c42386cabd0/src/BootstrapTable.js#L333

Any ideas?

src/BootstrapTable.js
```diff
@@ -142,7 +142,13 @@ class BootstrapTable extends Component {
...
 
     this.store.setData(nextProps.data.slice());
-    let page = options.page || this.state.currPage;
+
+    let page;
+    if (options.page !== 'undefined') {
+      page = options.page;
+    } else {
+      page = this.state.page;
+    }
```
scharf: why not `this.state.currPage`?

src/BootstrapTable.js
```diff
@@ -142,7 +142,13 @@ class BootstrapTable extends Component {
     const { options, selectRow } = nextProps;
 
     this.store.setData(nextProps.data.slice());
-    let page = options.page || this.state.currPage;
+
+    let page;
+    if (options.page !== 'undefined') {
```
scharf: you could also say `options.page != null'`. This will compare with `undefined` as well as `null`.

Note: do not use `!==` see [`null`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/null)

ianks: yeah thats how i originally did it, eslint was yelling at me so i thought this may have been a pattern in the code base. if @AllenFang agrees we can just adjust the `.eslintrc` to allow for double equal null checks.

src/BootstrapTable.js
ianks: Typo, thanks :+1:

src/BootstrapTable.js
scharf: You should use `typeof options.page !== 'undefined'`. Because the statement you use, does not do what you expect! the value `undefined` is never `===` the string `'undefined'`!

ianks: I changed this: https://github.com/AllenFang/react-bootstrap-table/pull/447/commits/98499b2d3de0149f27e85a75e753d948e25d32a7

AllenFang: Thanks, I'll review it :+1:

AllenFang:
@ianks, I found a bug in this patch, in the [example](https://github.com/ianks/react-bootstrap-table/blob/98499b2d3de0149f27e85a75e753d948e25d32a7/examples/js/pagination/custom-pagination-table.js) you modify.  You can redirect to last page will cause an error, maybe the calculation on page size or total page is wrong.  Did you want to fix this bug or just let me fix it?

Thanks 👍

ianks:
@AllenFang I am having some trouble recreating the bug? do you want to just give it a whirl and either PR the patch to me or post it here?

AllenFang:
@ianks, I'll checkout your branch and fix it on my local, I've fixed it and will release on next version. The bug is in [example](https://github.com/ianks/react-bootstrap-table/blob/98499b2d3de0149f27e85a75e753d948e25d32a7/examples/js/pagination/custom-pagination-table.js) and you just click ```last``` button to redirect to last page and you will get the error on browser console.  Actually the problem is the wrong calculated last page, because the start page is dynamically currently.

Anyway, thanks your contribution, I'll merge this PR with my patch in these day.

ianks:
@AllenFang thank you!

AllenFang:
Released on ```v2.3.2```. Thanks you again :)

PR #463, opened 2016-05-18
Don't set column width when it is null

hkaya:
The current implementation results is the following warning:

Warning: `NaN` is an invalid value for the `width` css style property. Check the render method of `TableBody`.


AllenFang:
@hkaya, I'll check this out :)

AllenFang:
Released on ```v2.3.3```. Thanks your contribution. 💯

hkaya:
@AllenFang Thank you!

PR #472, opened 2016-05-21
Add mouse click event to save button

daiwhea:
I use {blurToSave: false} and { type: 'textarea' } to create a multiple line textaera, when I click the save button, it won't save changes at all.

So I add some lines to TableEditColumn.js handleKeyPress().

### DIFF
--- a/src/TableEditColumn.js
+++ b/src/TableEditColumn.js
@@ -26,8 +26,8 @@ class TableEditColumn extends Component {
       this.props.completeEdit(
         null, this.props.rowIndex, this.props.colIndex);
     } else if ('click' === e.type) {
-      var value = _this.refs.inputRef.value;
-      _this.props.completeEdit(value, _this.props.rowIndex, _this.props.colIndex);
+      var value = this.refs.inputRef.value;
+      this.props.completeEdit(value, this.props.rowIndex, this.props.colIndex);
     }
   }
 


src/TableEditColumn.js
```diff
@@ -25,6 +25,9 @@ class TableEditColumn extends Component {
     } else if (e.keyCode === 27) {
       this.props.completeEdit(
         null, this.props.rowIndex, this.props.colIndex);
+    } else if ('click' === e.type) {
+      var value = _this.refs.inputRef.value;
```
AllenFang: where's the ```_this```. @daiwhea, did you test it?

daiwhea:
OMG, I have made a mistake. I just modify the code in my project's node_modules/react-bootstrap-table. It should be just this in src. _this is the dist version which compiled by babel.

      else if ('click' === e.type) {
        var value = this.refs.inputRef.value;
        this.props.completeEdit(value, this.props.rowIndex, this.props.colIndex);
      }

AllenFang:
@daiwhea, exactly, you can do a patch again in this PR. BTW, I've a question:


> I use {blurToSave: false} and { type: 'textarea' } to create a multiple line textaera, when I click the save button, it won't save changes at all.

I'm not sure what's means, did you try to fix a bug? or a improvement?
sorry, I've no time to review and run your PR currently, so just ask you :)

daiwhea:
Yes, it's just a bug fix. When I use a textarea to edit large text, the save button doesn't work at all. I can only trigger saving by setting blurToSave = true. But I do want to trigger saving by click save button.

I think it's a bug.

AllenFang:
> When I use a textarea to edit large text, the save button doesn't work at all.

it seems like a bug, only happened on large content in textarea?

daiwhea:
Yes, it won't work in textarea, no matter how many content it is. 

In fact, I haven't seen save button for single line text, so I can only set blurToSave = true or type ENTER key to trigger save event.

AllenFang:
> Yes, it won't work in textarea, no matter how many content it is.

I'll check this out and what's bootstrap and browser version you used?



> I haven't seen save button for single line text

you means the input text field? if yes, there's no save button, you need to press ENTER to save content, that's why textarea need a save button for saving the content.

daiwhea:
bootstrap 0.14.8, chrome on mac osx

AllenFang:
> bootstrap 0.14.8

you means ```react-bootstrap@0.14.8```

daiwhea:
sorry, detail is : 

bootstrap 3.3.6
react-bootstrap 0.28.5
**react 0.14.8**
react-bootstrap-table 2.0.3

Maybe I should upgrade to react-bootstrap-table 2.3.x

AllenFang:
@daiwhea, yes, try to upgrade ```react-bootstrap-table```

daiwhea:
It doesn't work after upgrade. May you have a test?

AllenFang:
@daiwhea, I'll check this out!

AllenFang:
@daiwhea, could you please show me your codes? because I cant reproduce this issue, i work for me.

daiwhea:
It's a big project, and it's not public. So I just show some key codes. I think maybe the react@0.14.8 version I'm using is the main reason. I cannot upgrade to react@15 for some broken changes.

Do you have a public demo ?

`      var ths = ... // TableHeaderColumn defines
       var cellEditProp = {};
       if (!xs) {
          cellEditProp = {
            mode: "click",
            blurToSave: false
          };
        }

        var list = (<BootstrapTable cellEdit={cellEditProp} data={listData} striped={true} hover={true}
                                condensed={true}>{ths}</BootstrapTable>)`

AllenFang:
@daiwhea, ok, I'll try to use react@0.14.x again.

AllenFang:
@daiwhea, I've fixed this issue on ```v2.3.4```. Anyway, thanks your contribution 👍

PR #476, opened 2016-05-23
Made glyphicons for toolbar buttons overridable

snowzach:
This allows you to override both the text and the icons for for the export, insert and delete buttons.


AllenFang:
Thanks I'll review it

AllenFang:
@snowzach, any other solution, because your solution will cause the other user lose the glyphicon if I merge this PR.

PR #490, opened 2016-06-04
fix selected row className

### LINKED ISSUE #488: error in selectRow className
I believe this is a bug. Say I have my table with `trClassName="foo"` and `selectRow={{ className: "bar", ... }}` props. As it's clear from [this code](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/TableRow.js#L56), when the row is selected, the class parameter will be `foobar`, instead of `foo bar` which is what I would expect. A workaround is to do `selectRow={{ className: "bar ", ... }}` but it feels kind of hacky. I can work on a fix if you agree

itajaja:
close #488


AllenFang:
Released on ```v2.3.4```. Thanks your contribution

itajaja:
a pleasure!

Issue #506, opened 2016-06-15
Impossible to filter in null values


jfremy:
The fix for #427 (#428) is too aggressive in the sense that it prevents altogether to filter in a value of null.
Whatever the input in the filter input, it's impossible to include the line (it returns false systematically to the filter function on a null value)

I'd like to suggest a different fix. Instead of returning false, we set targetVal to the empty string in case the value is null or undefined. I believe this makes sense since the input from the user will be mostly text based (we could imagine custom filters for which it's not true) but in most cases it is. So assuming that having a value of null or undefined maps to an empty string seems sensible.

With that change, it becomes possible to filter in null values using the regex filter for instance and specifying a regex that only allows an empty string.

Pull request to follow

AllenFang:
closed by #507

PR #428, opened 2016-04-28
fix #427

Kurtas:
FIX [#427](https://github.com/AllenFang/react-bootstrap-table/issues/427) The filters doesn't work when the column contains null values

Issue #427, opened 2016-04-28
The filters doesn't work when the column contains null values


Kurtas:
Example of data 

`[
{
id: 1,
data: 'test'
},
{
id: 2,
data: null
},
{
id: 3,
data: 'test 2'
}
]`

AllenFang:
I'll check this out. Thanks @Kurtas

amitkumarsingh:
Hi Allen,

There are some objects where there are null values and this is the error i get on searchinh that particular field.
react-bootstrap-table.min.js:8 Uncaught TypeError: Cannot read property 'toString' of undefined.

AllenFang:
@amitkumarsingh, ok could you please open another issues :)  Thanks!

jfremy:
Hi guys, returning false for the filtering on null value might be too strong. It means you cannot filter in the values that are null.
What do you think about rather replacing null with the empty string, that way, there would still be way to write a filter that includes those lines (like the Regex filter)

AllenFang:
#443, actually, I cant reproduce this issue. so just pending it.. BTW, thanks @jfremy

PR #507, opened 2016-06-15
Fix for #506 impossible to filter in null values

jfremy:
The fix for #427 (#428) is too aggressive in the sense that it prevents altogether to filter in a value of null.
Whatever the input in the filter input, it's impossible to include the line (it returns false systematically to the filter function on a null value)

I'd like to suggest a different fix. Instead of returning false, we set targetVal to the empty string in case the value is null or undefined. I believe this makes sense since the input from the user will be mostly text based (we could imagine custom filters for which it's not true) but in most cases it is. So assuming that having a value of null or undefined maps to an empty string seems sensible.

With that change, it becomes possible to filter in null values using the regex filter for instance and specifying a regex that only allows an empty string.

### DIFF
--- a/src/store/TableDataStore.js
+++ b/src/store/TableDataStore.js
@@ -387,7 +387,9 @@ export class TableDataStore {
       let filterVal;
       for (const key in filterObj) {
         let targetVal = row[key];
-        if (targetVal === null) return false;
+        if (targetVal === null || targetVal === undefined) {
+          targetVal = '';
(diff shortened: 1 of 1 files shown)


AllenFang:
@jfremy, I think it's good for fixing this issues :)

AllenFang:
@jfremy, released on ```v2.3.6```. Thanks your contributions :)

PR #508, opened 2016-06-15
Follow up on custom data accessor #418

jfremy:
Hi,
I've taken #418 by @jdeniau and updated it to work on the latest master.
I've also updated the code so that search works with the custom data accessor and updated the example to show that.

That leave editing but I'm not sure you'd want to mix the 2. Let me know what you think


AllenFang:
@jfremy, thanks your contribution, I'll review it in later. Anyway, I want to fix this both issues but actually, I hope to have a better solution to cover all the condition for nest data. I've some plans but just no time to do...

welcoMattic:
@jfremy Hi! Will you fix your PR or can I handle it? This feature looks like very important for usability of the component, may be we should fix it to publish a new release? cc @AllenFang

AllenFang:
Actually, It's hard to finish this job, because there're too many place which access the data field, so these place you need to consider how to apply the nest data structure on that.

I've a idea at before, but no time to make it to true...

Following is my plan, 

Keep in mind, we should try to avoid touch the source code too many, maybe we can provide a **meta transfer** which responsibility is **flat** the nested data then we just give the flatten data into ```<BootstrapTable>``` 

for example,  the nested data is 
```js
[
{ id: 0, person: { name: 'Allen', age: 27 }, city: { address: 'xxx', region: 'ooo' } },
.....
]
```
We should transfer out the following result by this nested data according the user spec via our **meta transfer**
```js
[
{ id: 0, name: 'Allen', age: 27, address: 'xxx', region: 'ooo' }
.....
]
```

But there's something we need to care, like ```TableDataStore.js``` and some exposed API, if anyone got interesting, feel free to connect me, maybe this is a better way to follow :)

Thanks :)

Siyfion:
Not sure I agree with your idea there @AllenFang, for example what if the nested data looked like:
```
[
  { id: 0, person: { name: 'Allen', age: 27 }, spouse: { name: 'Tina', age: 24 } },
  ...
]
```
or similar?

Surely the sensible way is to change the code to use something like `_.get()` and `_.set()` from lodash?

Mithun-UA (approved):

PR #554, opened 2016-07-08
Footer

r1g0:
Hello, I added a footer.  To show it there needs to be set a `showFooter={ true }` as props.  

When `showFooter` is true it shows the same text in the header in the footer unless a `footerText='something'` is set and then it preffers that one.

I created a TableFooter from the TableHeader and removed the sorting and filtering components.

And also added an example.

### DIFF
--- a/src/TableFooter.js
+++ b/src/TableFooter.js
@@ -1,5 +1,7 @@
 import React, { Component, PropTypes } from 'react';
 import classSet from 'classnames';
+import Const from './Const';
+import SelectRowHeaderColumn from './SelectRowHeaderColumn';
 
 class TableHeader extends Component {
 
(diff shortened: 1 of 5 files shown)


AllenFang:
@r1g0, thanks :) I'll check it out.  If it's worth to merge, I'll ask you merge into ```v3.0.0-dev``` branch

AllenFang:
Ref #497, dont forget to review this PR

src/TableFooter.js
```diff
@@ -0,0 +1,46 @@
+import React, { Component, PropTypes } from 'react';
+import classSet from 'classnames';
+
+class TableHeader extends Component {
```
anujemc: If its a tablefooter than class name should be `TableFooter` rather than `TableHeader`

PR #565, opened 2016-07-16
improvement react-modal esc behavior

madeinfree:
Hi, @AllenFang 

I added the esc (close) behavior in ToolBar react-modal, and I would not to mutate the origin "modal" variant by just use React.cloneElement, I think maybe it can be friendly for user to switch it.

### DIFF
--- a/src/toolbar/ToolBar.js
+++ b/src/toolbar/ToolBar.js
@@ -109,6 +109,12 @@ class ToolBar extends Component {
     this.setState({ isInsertModalOpen: false });
   }
 
+  handleModalRequestClose = (modal) => {
+    return React.cloneElement(modal, {
+      onRequestClose: this.handleModalClose
+    });
+  }
+
   handleShowOnlyToggle = () => {
     this.setState({
       showSelected: !this.state.showSelected
@@ -218,9 +224,7 @@ class ToolBar extends Component {
     let modal = this.props.enableInsert ? this.renderInsertRowModal() : null;
 
     if (this.props.enableInsert) {
-      modal = React.cloneElement(modal, {
-        onRequestClose: this.handleModalClose
-      });
+      modal = this.handleModalRequestClose(modal);
     }
 
     return (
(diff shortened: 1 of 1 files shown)


src/toolbar/ToolBar.js
```diff
@@ -215,7 +215,13 @@ class ToolBar extends Component {
     }
 
     const searchTextInput = this.renderSearchPanel();
-    const modal = this.props.enableInsert ? this.renderInsertRowModal() : null;
+    let modal = this.props.enableInsert ? this.renderInsertRowModal() : null;
+
+    if (this.props.enableInsert) {
```
madeinfree: I think chang this condition can help user to switch the esc close behavior.

src/toolbar/ToolBar.js
```diff
@@ -218,9 +224,7 @@ class ToolBar extends Component {
     let modal = this.props.enableInsert ? this.renderInsertRowModal() : null;
 
     if (this.props.enableInsert) {
-      modal = React.cloneElement(modal, {
-        onRequestClose: this.handleModalClose
-      });
+      modal = this.handleModalRequestClose(modal);
```
madeinfree: When use the condition to switch it, maybe didn't necessary to named "modal"

AllenFang:
@madeinfree, cool, I'll review it. Thanks your 👍

src/toolbar/ToolBar.js
AllenFang: I dont know what's your means

AllenFang:
> I added the esc (close) behavior in ToolBar react-modal

After review, I found you attach a ```onRequestClose``` event on modal? but I've no idea about how to connect it with click esc.

src/toolbar/ToolBar.js
AllenFang: we just follow the bootstrap, esc will trigger close event for modal at default. If I've free time, I'll make it as configurable.

Issue #206, opened 2016-01-10
Row selection (clickToSelect) slow with long tables.


finkelm:
Noticeable lag appears on long lists between clicking and seeing the highlight of row.

vgoklani:
same issue here

AllenFang:
I'll find this problem quickly :) Thanks your feedbacks

vgoklani:
The search box is also very slow on long lists. Are you debouncing the queries? The search box is actually unusable with 1500 rows.

AllenFang:
the search will search all table rows and each cells, I'll try to find some debouncing in the queries

finkelm:
The only alternative is pagination. It might be all the listeners being created?

AllenFang:
@finkelm, I've doing some tuning on branch, will release in these day :)

AllenFang:
@finkelm, I've improve the performance for this case, check the ```v1.4.2```
@vgoklani, because search text will trigger a search when keyup, so I suggest you customize a search input on you component and calling ```handleSearch``` instead of using built-in search

tell me if no any issues :)

TMandzu:
i have the same problem. clickToSelect needs about 2-3 secs to select on 3K rows. Is it really fixed?

AllenFang:
Not yet,,, sorry, current focus on `v3.0.0` development, I'll do it after this release.

TMandzu:
all right, thanks a lot :)

gary-menzel:
I'm getting this only on just under 400 rows.  But, after I apply a filter of some sort, the issue is not noticeable. Looking forward to the fix.

MichalRemis:
Some progress on this? I have same issue on 3.4.4.. it takes seconds on a big table to select a row

gyerzh:
the same for me

PR #566, opened 2016-07-16
Improvement table body row render calculator

madeinfree:
fixed issue #206 

This PR try to improve the table row render for big data render and select (checkbox, radio) click, now table row selection be faster than before.

I added the empty table row for invisible data when it couldn't show in the table to avoid the TableRow render do.

But I can't promise it not affect the other feature, it just an experimental maybe we have to discuss it.

try it :)

### DIFF
--- a/examples/js/selection/big-row-select-table.js
+++ b/examples/js/selection/big-row-select-table.js
@@ -1,0 +1,36 @@
+/* eslint max-len: 0 */
+import React from 'react';
+import { BootstrapTable, TableHeaderColumn } from 'react-bootstrap-table';
+
+
+const products = [];
+
+function addProducts(quantity) {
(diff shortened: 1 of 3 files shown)


AllenFang:
hi @madeinfree, it's a awesome idea for rendering big rows, I'll check it out and it is necessary to do more test. Thanks a lots

jvallarvisp:
Hi @AllenFang, I think one of the commits on this can be implemented to infinite scrolling.

AllenFang:
@jvallarvisp, which one? BTW, I would like to merge this into ```v3.0.0-dev```, because I think it can be a new feature on next major version upgrade, so maybe you need to wait ```v3.0.0``` release.

PR #579, opened 2016-07-22
Fixed #578

### LINKED ISSUE #578: Feature Request: Accept height/maxHeight of type 'number'
For some reason, the `height` and `maxHeight` properties of `BootstrapTable` only accept strings.  This is documented [in the table's `propTypes`](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/BootstrapTable.js#L907) so it's not a bug, but I was surprised when passing a number [threw an uncaught exception here](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/BootstrapTable.js#L875).

It would be nice if these fields accepted a number, and implicitly treated them as a pixel value [the same way the `style` prop does](https://facebook.github.io/react/tips/style-props-value-px.html).

I want this to work:
```jsx
const MAX_TABLE_HEIGHT = 500;
<BootstrapTable height={MAX_TABLE_HEIGHT}/>
```

Right now I have to do this:
```jsx
const MAX_TABLE_HEIGHT = 500;
<BootstrapTable height={MAX_TABLE_HEIGHT+'px'}/>
```

madeinfree:
fixed #578 issue,

Intent BootstrapsTable allow string and number.

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -912,7 +912,7 @@ class BootstrapTable extends Component {
 BootstrapTable.propTypes = {
   keyField: PropTypes.string,
   height: PropTypes.oneOfType([ PropTypes.string, PropTypes.number ]),
-  maxHeight: PropTypes.string,
+  maxHeight: PropTypes.oneOfType([ PropTypes.string, PropTypes.number ]),
   data: PropTypes.oneOfType([ PropTypes.array, PropTypes.object ]),
   remote: PropTypes.bool, // remote data, default is false
   striped: PropTypes.bool,
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -810,7 +810,12 @@ class BootstrapTable extends Component {
             onDropRow={ this.handleDropRow }
             onSearch={ this.handleSearch }
             onExportCSV={ this.handleExportCSV }
-            onShowOnlySelected={ this.handleShowOnlySelected }/>
+            onShowOnlySelected={ this.handleShowOnlySelected }
+            insertModalHeader={ this.props.options.insertModalHeader }
+            insertModalFooter={ this.props.options.insertModalFooter }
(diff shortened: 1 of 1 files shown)


src/BootstrapTable.js
```diff
@@ -872,6 +877,8 @@ class BootstrapTable extends Component {
   }
 
   _adjustHeight = () => {
+    if (typeof this.props.height === 'number' && !isNaN(this.props.height)) return this.props.height;
```
islemaster: Thanks for grabbing this feature request so fast!  I'm don't think this change works though.  Nothing uses the return value of `_adjustHeight` - instead, the method directly modifies the container height to be the requested height less the header height, unless the requested height is specified as a percentage.  (There's an existing bug here, in that it doesn't read `em` values, etc. correctly, but I don't think that's a priority.)  Would it be better to just do something like this?
```jsx
const stringHeight = String(this.props.height);
if (stringHeight.indexOf('%') === -1) {
  this.refs.body.refs.container.style.height = parseFloat(stringHeight, 10) - this.refs.header.refs.container.offsetHeight + 'px';
}
```

madeinfree:
Hi, @istanishev Thanks for your suggestion, I'll check it later.

src/BootstrapTable.js
```diff
@@ -810,7 +810,12 @@ class BootstrapTable extends Component {
             onDropRow={ this.handleDropRow }
             onSearch={ this.handleSearch }
             onExportCSV={ this.handleExportCSV }
-            onShowOnlySelected={ this.handleShowOnlySelected }/>
+            onShowOnlySelected={ this.handleShowOnlySelected }
+            insertModalHeader={ this.props.options.insertModalHeader }
```
AllenFang: @madeinfree, I thinks you got the code from ``v3.0.0-dev``` into this branch. But that's ok, I'll remove it if this PR is ok :)

PR #582, opened 2016-07-23
Add paginationPosition option

garlab:
This option allows to set the position of the pagination bar:
- 'top' will position it on top, between the table and the toolbar
- 'bottom' (default) will position it below the table (current behaviour).
- 'both' will display two pagination bar, one on top and the other on the bottom.

PR #635, opened 2016-08-16
Add paginationPosition option

garlab:
Hello,

This is the same as #582 but in v3.0.0-dev; I also squashed the two commits.

This option allows to set the position of the pagination bar:
- 'top' will position it on top, between the table and the toolbar
- 'bottom' (default) will position it below the table (current behaviour).
- 'both' will display two pagination bar, one on top and the other on the bottom.


AllenFang:
@garlab  thanks you :)

PR #646, opened 2016-08-19
use return value from onCellEdit as new cell value

### LINKED ISSUE #631: Add filter/conversion function that can be applied to editable cells
**Version:** 2.4.2

`beforeSaveCell` or `onCellEdit` should be allowed to return a new value that is then used for `newVal`.

bontibon:
fixes #631


AllenFang:
@bontibon, thanks your contribution :)

PR #661, opened 2016-08-26
added some options for behaviors in PaginationList

swearer23:
1. options.alwaysShowAllBtns added,  which default value is false agrees with the behaviors of current version, if it's set to true , prev-page and next-page should be always showing

2. options.withFirstAndLast added,  which default value is true agrees with the behaviors of current version, if it's set to false, first-page and last-page should not be added during PaginationList::getPages method, which result in them not being displayed

BTW: why not ignore dist foler?


AllenFang:
@swearer23, thanks your contributions 👍

AllenFang:
#497

Issue #659, opened 2016-08-25
update sizePerPage through componentWillReceiveProps does not update select list


Raffiki:
I have the following table. 

`SizePerPage` defaults to `10` but should be updated through `componentWillReceiveProps`.

If this happens and `options.sizePerTable` becomes say `30`, 30 rows are indeed rendered, but the `sizePerPage` widget still shows `10`.
[of interest: on the next `render`, by opening another component and coming back to this one, the widget updates to the desired `30`]

Is there anything obvious I am overlooking here? Do I pass the correct parameters?

```
export default class PaginationTable extends React.Component {
  constructor (props) {
    super(props);
    this.state = {
      options: {
        page: this.props.data.options._page || 1,  // which page you want to show as default
        onPageChange: this.props.onPageChange,
        onFilterChange: this.props.onFilterChange,
        onRowClick: this.props.onRowClick,
        sizePerPage: this.props.data.options._pageSize || 10,  // which size per page you want to locate as default
        pageStartIndex: 1, // where to start counting the pages
        paginationSize: 9,  // the pagination bar size.
        prePage: '<', // Previous page button text
        nextPage: '>', // Next page button text
        firstPage: '<<', // First page button text
        lastPage: '>>' // Last page button text
      }
    };
  }

  componentWillReceiveProps (newProps) {
    const sizePerPage = _.get(newProps, 'data.options._pageSize');
    if (sizePerPage !== this.props.data.options._pageSize) {
      this.setState({
        options: _.assign({}, this.state.options, {sizePerPage})
      });
    }
  }

  render () {
    return (
      <BootstrapTable
        data={this.props.data.data}
        fetchInfo={{ dataTotalSize: this.props.data.numRows }}
        remote={true}
        pagination={true}
        options={this.state.options}
        striped={true}
        hover={true}
        selectRow={this.props.selectRow}>
        {this.props.children}
      </BootstrapTable>
    );
  }
```

Raffiki:
It is on the next `mount` (vs. render) of the component, that the correct `sizePerPage` is shown. So my 2cnts is `sizePerPage` is not picked up at `componentWillReceiveProps` on the react-bootstrap-table side. Need to doublecheck this though

Raffiki:
adding `sizePerPage` to state when `remote` is true seems to fix this.
https://github.com/AllenFang/react-bootstrap-table/blob/master/src/BootstrapTable.js#L173
Any reason why this is explicitly left out?

[...]

PR #672, opened 2016-08-31
Pagination fix sizePerPage for issue #659

doshisunny:
Updated to use sizePerPage from remote component when remote = true.


AllenFang:
I'll merged it after testing. Thanks you guys :)

PR #692, opened 2016-09-08
Add "defaultSearch" option in documentation

syndbg:
BootstrapTable accepts "defaultSearch" option that's not mentioned in the docs


AllenFang:
@syndbg, thanks your contribution :)

Issue #683, opened 2016-09-05
Bootstrap 4


razagill:
Hey guys,
Trying it out with bootstrap 4 but looks like the pagination doesn't have the correct CSS. They are just appearing as a list. Anyone else experiencing it?

AllenFang:
@razagill, I haven't yet to test with bootstrap 4. Maybe we need a good plan to do this job :)  But before this, I need to get ```v3.0.0``` done in near future.

razagill:
@AllenFang cool. I noticed there are some new classes that Bootstrap 4 adds to the pagination ```ul``` and ```li``` elements. I added those classes and it's working fine for now. Maybe we can have a small conditional to check if the version of bootstrap is 4 then add those classes.

AllenFang:
@razagill, PR is welcome :)

philohelp:
That would be a great help, thanks !

philohelp:
@razagill Meanwhile would you mind releasing the trick concerning bs4 new classes ? Thanks !

razagill:
@philohelp Sure, here you go: https://github.com/razagill/react-bootstrap-table/blob/master/src/pagination/PageButton.js

AllenFang:
Fixed on ```v2.5.4```. Thanks :)

PR #700, opened 2016-09-12
Added bootstrapv4 classes to pagination

razagill:
Solves #683


razagill:
@AllenFang: I haven't found a good way to differentiate between bootstrap versions as of now. There are some people who are using the comments in the CSS/JS files for that but I'm guessing most of the react users have some building process (webpack, browserify) which will remove those comments, so it's not reliable to use that. The current option is just to add these classes as is, obviously they would not have any effect for bootstrapv3.

Issue #460, opened 2016-05-17
Grouping rows


lioralt:
Hi,
Is it possible to dynamically group rows by a specific column ?
Also, is it possible to expand/collapse rows with react-bootstrap-table ?

thanks.

AllenFang:
> Is it possible to dynamically group rows by a specific column ?

I'm not sure what's that, so any description or examples?

BTW, expand row has been wished on #362. Anyway, I'm busy on my work currently, so just can do some little fixing or patching, sorry about that I can't support these useful feature in near future. Maybe I'll do it if there has some free time, so PR is welcome :)

AllenFang:
in ```v2.6.0```, we support expand row, check [example](https://github.com/AllenFang/react-bootstrap-table/blob/master/examples/js/expandRow/expandRow.js). We will update document in these day

Issue #362, opened 2016-03-26
Expandable Rows


sugargreenbean:
Looking for a way to click on a row or an icon within a column and then it expands with more information about that row. Something similar to this: http://daviferreira.github.io/react-sanfona/ or this http://mandarinconlabarba.github.io/react-tree-menu/example/index.html

As the table seems to only expose columns, I have no idea how to customize rows. Thank you for any help!

madeinfree:
Hi, @sugargreenbean I discussion this feature with @AllenFang recently, it sound a good idea,but we have to observation and plan this feature. maybe we can achieve it.

AllenFang:
@sugargreenbean, I thinks it's a nice feature for ```react-bootstrap-table```, but it actually need some time to do it. Hope we can support this feature in near future

sugargreenbean:
Ok, good to hear! Meanwhile, I am working on this. I see that the rows are generated as <tr className="foo" data-reactid="..."> Where is this being generated? I thought it was in TableRow.js, but can't seem to change what is in the tr tag, as in I want the tag to say <tr ClassName="foo" data-reactid="..." data-toggle="collapse">

AllenFang:
@sugargreenbean, yes, it is in ```TableRow.js```, I think it can be changed by some flag, for a roughly example: 
```
const content = this.props.enableExpand ? this.renderExpandContent() : this.props.children;
```
However, I think expand row is a little hard to implement, maybe we can discuss more on it.

sugargreenbean:
This is what I have currently tried:
Following this: http://www.w3schools.com/bootstrap/bootstrap_ref_js_collapse.asp
I am trying to add to a current selected row tr tag:
`data-toggle="collapse" data-target="#foo"`
And then be able to add a row underneath current selected row with handleAddRow. Then I want to customize this Row to not look like a row and populate with extra info about the row.

sugargreenbean:
Sorry, I updated my 2nd comment with proper styling so you can see my code examples which were not previously viewable.

AllenFang:
```data-reactid ``` is generated by React, ```className``` is generated by ```trCss```, you can check [here](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/TableRow.js#L54).

I thinks you can just render ```data-toggle="collapse" data-target="#foo"``` by some flag. Hope I answer your question. BTW, what's means for *And then be able to add a row underneath current selected row with handleAddRow*.
[...]

PR #721, opened 2016-09-21
Sections support

kwent:
Following those issues https://github.com/AllenFang/react-bootstrap-table/issues/460, https://github.com/AllenFang/react-bootstrap-table/issues/362.

I hacked the library to support a **plain JS object** ( with a deep of 1 ) in addition of an **array** resulting the table to cut **sections** (multiple **tbody**) for each keys.

Final result is:

screen shot 2016-09-21 at 4 21 45 pm

**Data format** is now:

``` js
const products = { categoryA: [ { id: 1, name: 'Item 1', price: 1000 },
                                { id: 2, name: 'Item 2', price: 2000 }
                               ],
                   categoryB: [ { id: 3, name: 'Item 3', price: 3000 },
                                { id: 4, name: 'Item 4', price: 4000 },
                                { id: 5, name: 'Item 5', price: 5000 }
                               ]
                 };
```

Demo file is here : https://github.com/kwent/react-bootstrap-table/blob/sections/examples/js/sections/demo.js

It's only tested with my use case. (Simple table with cell edit enable without breaking current features).

**I don't think it's ready to merge** but i think we should keep improve this PR to support all feature around a JS object. (checkbox untested, sort untested etc...)


AllenFang:
That's cool, I'll check this out :)

Issue #690, opened 2016-09-08
Issue with insert row modal


edulis8:
Hello, thanks for your attention so far.

I am having an issue with the insert row modal. When I click "New" the modal appears, but the entire screen appears darkened, and as soon as I click on a field on the modal, it disappears again.

I have included a screenshot.

I would also like to only allow the user to enter values for 2 columns (in my case, item# and quantity), not all the columns. After the user enters values for 2 columns, my app will source the other columns dynamically, so they don't need to enter anything else. Is there a way to select which columns the user can enter during a row insertion?

![screen shot 2016-09-07 at 11 24 19 pm](https://cloud.githubusercontent.com/assets/13525195/18339091/53bed8c4-7552-11e6-9bfa-4f16d4a2d076.png)

AllenFang:
HI @edulis8, firstly, the issue of modal being darkened is a known issue, maybe cause by different bootstrap version(there's someone encounter this issue). The second one, we only allow you to configure the editable for column instead of hiding it. 

Anyway, this two issue will be solved on ```v3.0.0``` in near future(I'm not very sure what time I can finish it). In ```v3.0.0```, I drop bootstrap modal, and use ```react-modal``` so that avoid these annoying  bugs and we will allow user to customize the everything in modal, so you can control the which field should be display on modal.

Feel free to discuss with me

Thanks :)

AllenFang:
BTW, you can check #497

edulis8:
What bootstrap version should I use do you think?

Eric

AllenFang:
@edulis8, try ```3.3.2```

hustshawn:
So react-strap-table should depend on bootstrap? or react-bootstrap ?

AllenFang:
Depend on bootstrap, you should include the ```bootstrap.js``` and ```jquery```. Anyway, like I said, I'll remove these dependencies in the ```v3.0.0```

hustshawn:
@AllenFang Did you fixed in v3.0.0, I cannot find the related update on that branch. Also I investigated via react dev tools, find the modal seems rendered but not appear on the top. Can it be the root cause?

AllenFang:
@hustshawn, I'm still fixing the dependency issue on my local, almost done, but just not yet push back to repo. And currently, I've abandon the native bootstrap modal and switch to ```react-modal```(https://github.com/AllenFang/react-bootstrap-table/blob/v3.0.0-dev/package.json#L68)



[...]

PR #745, opened 2016-09-30
Modified ways to trigger insert modal

hustshawn:
Referred to the Bootstrap official [demo,](http://getbootstrap.com/javascript/#live-demo) find it might be use the element id to trigger the modal with button.
Just made small changes to fix the issue. Please review and comment. Thanks.


AllenFang:
@hustshawn, thanks your contribution, I'll review it, but this PR can solve the #690 ?

because ```data-target``` can accept a class or id, so I'm not very clear why this changed?

Issue #430, opened 2016-04-29
Printing the table makes the header columns width different from the data


ehelguero:
Hello! I´m am printing the table and it gets like this:

<img width="809" alt="captura de pantalla 2016-04-29 a las 6 20 46" src="https://cloud.githubusercontent.com/assets/1289195/14908685/4110f6d6-0dd2-11e6-9440-3b080d8cde97.png">

Is there a way to fix it? With the new version?

Thanks

AllenFang:
> printing the table

what's that means?, because I feel the the table in the picture is not a react-bootstrap-table``` 

could you please give me the example? and what's browser and ```react-bootstrap-table``` you used?

Thanks :)

ehelguero:
Hello, i am using window.print to print out the screen. Thats how it looks when i save the react-bootstrap-table to a PDF.

AllenFang:
@ehelguero, ok, I see. I can't figure out what happen on window.print with react-bootstrap-table. I'll do some survey, but not sure I can solve it.

ehelguero:
Ok, thank you very much in advance.

ibilanchuk:
Hi, i have same problem 
Will be very glad if you'll fix it.
Thanks.

AllenFang:
Try ```v2.5.6```.  Feel free to discuss with me if problem still there

nathanpjones:
This resolves the issue _unless_ you set `width` for any column. If you set `width`, then the headers are misaligned in both screen and print.

PR #762, opened 2016-10-06
Add an option to make the tables printer friendly. Fixes issue #430

mark-riley:
This might not be optimal. If you just want the table headers to be the same size as the columns, it is fine. If you need scroll bars on your table, then it won't work. The two are mutually exclusive.

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -892,7 +892,8 @@ class BootstrapTable extends Component {
   }
 
   _adjustTable = () => {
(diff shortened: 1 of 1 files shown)


AllenFang:
@mark-riley, thanks your contribution. could you confirm this solution will fix the unalign when printing?

mark-riley:
I can confirm that it fixes it on my set up (Ubuntu, Firefox.)

AllenFang:
@mark-riley, released on ```v2.5.6```. Thanks

PR #772, opened 2016-10-13
fixed #768 to filter number 0

### LINKED ISSUE #768: Search fails on number 0
When binding to a column with a value of 0 and then typing 0 in the search box doesn't properly return the intended results.

TableDataStore.js

line 489

if (this.colInfos[key] && row[key]) {

This returns 0 (which is false)

Converting the data first to a string gets around the issue, but not the desired solution.

madeinfree:
fixed #768, This is my solution to avoid filter misunderstand 0 become false.

```
<BootstrapTable data={ products } search={ true } options={ options }>
  <TableHeaderColumn dataField='name' searchable={ false }>Fruit Name</TableHeaderColumn>
  <TableHeaderColumn dataField='price' searchable={ false }>Product Price</TableHeaderColumn>
</BootstrapTable>
```

when id equals number 0 then filter can know it in this special case.

### DIFF
--- a/src/store/TableDataStore.js
+++ b/src/store/TableDataStore.js
@@ -486,7 +486,11 @@ export class TableDataStore {
       // http://jsperf.com/for-vs-foreach/66
       for (let i = 0, keysLength = keys.length; i < keysLength; i++) {
         const key = keys[i];
-        if (this.colInfos[key] && row[key]) {
+        // fixed data filter when misunderstand 0 is false
+        let filterSpecialNum = false;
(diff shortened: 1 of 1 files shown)


AllenFang:
@madeinfree, thanks, busy on work, I'll also think how to solve this issue. Anyway, thanks your contribution!

AllenFang:
Will be released on next version. Thanks your contribution 💯

AllenFang:
Released on ```v2.5.6```. Thanks

PR #777, opened 2016-10-16
Add onRowDoubleClick option to BootstrapTable

### LINKED ISSUE #250: Doubleclick event handler support
First, thanks for your work and for this library!

The default "insert new row" feature is nice, but in many situations, it is better to allow the programmer to handle this in a specific way. For this, it would be useful if the programmer could provide an event handler for double clicking on a row.

jlinn:
This would resolve #250


AllenFang:
@jlinn, thanks for contribution. I'll check this out. :)

AllenFang:
Released on ```v2.5.6```. Thanks :0

PR #780, opened 2016-10-18
Font Awesome update

gforge:
Here's a version that uses the Font Awesome package instead of the Glyphicons that are deprecated in Bootstrap 4. 

There was no `export` alternative available so I used a [downward arrow](http://fontawesome.io/icon/arrow-circle-down/).


AllenFang:
@gforge, thanks for your contribution!  I'll spend some time to review it!!

PR #798, opened 2016-10-26
Enhancement: Export hidden columns

mph006:
Added exportHidden boolean which exports hidden columns to csv if flagged true.
Helps when screen real-estate is tight, but you still want to export all the underlying data.

exportHidden (bool) default:false


AllenFang:
@mph006, you can apply ```export={true}``` on ```TableHeaderColumn``` for exporting hidden column. I think this is what you want :)

mph006:
@AllenFang Yep...that did it. Somehow I missed that in the docs. Thanks!

AllenFang:
@mph006, that's ok, you did a great job. Thanks your contribution 👍

PR #822, opened 2016-11-11
Add Footers

sameersemna:
Add footers to bootstrap table and pass a prop for footer_data


package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "react-bootstrap-table",
-  "version": "2.5.8",
+  "name": "react-bootstrap-table-footer",
```
AllenFang: You should no change the ```name```, ```version```, ```url```, ```author``` in package.json

PR #843, opened 2016-11-17
Throw errors instead of string literals

maaaikoool:
Throwing literals can mess up the stack trace and produce weird bugs. I forgot to add an unique key and [redbox](https://github.com/commissure/redbox-react) got messed up and was quite hard to find the actual error.

http://eslint.org/docs/rules/no-throw-literal
http://www.devthought.com/2011/12/22/a-string-is-not-an-error/


AllenFang:
@maaaikoool, sorry for lately reply and thanks your contributions :)

PR #868, opened 2016-12-05
#0 getComputedStyle is not defined error

oalmali:
Hello;

I've been enjoying react-bootstrap-table for a while thanks for all effort.

While testing react-bootstrap-table with Enzyme a problem came up. When you mount a component that contains react-bootstrap-table it says getComputedStyle() is not defined. So that's why I've changed getComputedStyle() calls with window.getComputedStyle().

Thanks and regards.


AllenFang:
@oalmali, thanks your contribution. BTW, why adding lib folder?

oalmali:
@AllenFang I did not do it on purpose. I was trying to do a quick fix. So when you merge it you can remove it without any hesitation.

PR #914, opened 2016-12-27
Fixes #912

### LINKED ISSUE #912: onRowMouseOver called with wrong arguments when expandableRow is set
When using onRowMouseOver on a table with expandableRow set, incorrect arguments are passed to the function : on the first row it works ok, on the second row the third row object is passed, on the third row the fifth object is passed, and so on until the object passed to the callback become undefined.

I think it's caused by the addition of hidden rows when using expandableRow .

Code below is a modification of the basic example that reproduces the problem.

```js
const products = [];

function addProducts(quantity) {
  const startId = products.length;
  for (let i = 0; i < quantity; i++) {
    const id = startId + i;
    products.push({
      id: id,
      name: 'Item name ' + id,
      price: 2100 + i
    });
  }
}

addProducts(5);

const options = {
  onRowMouseOver: (row) => console.log(row)
};

export default class BasicTable extends React.Component {
  render() {
    return (
      <BootstrapTable
        expandableRow={ () => true }
        expandComponent={ () => 'Hello World' }
        data={ products }
        options={ options }>
          <TableHeaderColumn dataField='name'>Product Name</TableHeaderColumn>
          <TableHeaderColumn dataField='price'>Product Price</TableHeaderColumn>
      </BootstrapTable>
    );
  }
}
```

ptitFicus:
Not sure this works for every case, but it definitely fixes mine.

### DIFF
--- a/src/TableRow.js
+++ b/src/TableRow.js
@@ -39,7 +39,7 @@ class TableRow extends Component {
     if (e.target.tagName !== 'INPUT' &&
         e.target.tagName !== 'SELECT' &&
         e.target.tagName !== 'TEXTAREA') {
-      const rowIndex = e.currentTarget.rowIndex + 1;
+      const rowIndex = this.props.index;
       if (this.props.onRowDoubleClick) {
         this.props.onRowDoubleClick(rowIndex);
       }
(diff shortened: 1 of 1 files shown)


AllenFang:
@ptitFicus, thank your contribution, I'll check this out 👍

AllenFang:
Released on ```v2.8.1```, thank you

PR #919, opened 2016-12-29
Fix #918

### LINKED ISSUE #918: 2.8.1 broke sorting
After upgrading to v2.8.1, sort icons are not visible anymore and clicking on a header cell does nothing.

Examples of the sort section are broken too.

Rosaniline:
Sorry that I made two stupid mistakes that results in `rowCount` equals to `NaN`. I had it fixed and shall be fine right now, please check.
Besides, I wondered if it is necessary to add default value and prop type for `TableHeaderColumn.row`?
Thank you.

### DIFF
--- a/src/TableHeader.js
+++ b/src/TableHeader.js
@@ -32,8 +32,9 @@ class TableHeader extends Component {
       'table-condensed': this.props.condensed
     }, this.props.tableHeaderClass);
 
-    const rowCount = Math.max(React.Children.map(this.props.children, elm =>
-      Number(elm.props.row)));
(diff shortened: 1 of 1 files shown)


AllenFang:
@Rosaniline, never mind, make it as default value is good and ```TableHeaderColumn.row``` is started from ```v2.8.0``` by header column span feature, that's my fault to review that PR more detail.

AllenFang:
Thanks your patch 👍

AllenFang:
god.. I forgot to make zero as default value..QQ  I'll add it in next version..

PR #923, opened 2016-12-30
Allow percentage values in column width. Fixes #910

### LINKED ISSUE #910: Setting column width with percentage
Hi 
This is an improvement suggestion ! So that the columns widths will better adapt when resizing the application window.

By the way, how the column width is calculated by default ?

ptitFicus:
This modification allows to put values such as '50%' in the width property of TableHeaderColumn components.

I doubt this completely solve the issue since it's a very simplistic modification. However I haven't find any scenario when it doesn't work.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -63,11 +63,10 @@ export default {
         display: column.hidden ? 'none' : null
       };
       if (column.width) {
-        const width = parseInt(column.width, 10);
-        style.width = width;
+        style.width = column.width;
         /** add min-wdth to fix user assign column width
         not eq offsetWidth in large column table **/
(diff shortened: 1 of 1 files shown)


AllenFang:
LGTM

> I doubt this completely solve the issue since it's a very simplistic modification. However I haven't find any scenario when it doesn't work.

yap, the width of bootstrap table is a noisy problem.  I'll test it more detail. Thanks your contribution. 👍

AllenFang:
Close this PR cause of unknown repository, I'll open up if repo is back.

ptitFicus:
My bad, reopened #993.

Sorry about that

PR #933, opened 2017-01-04
Adding support for styled-components

jsangilve:
[Styled components](https://github.com/styled-components/styled-components) is a package to style components with _CSS in JS_ concepts. 

As explained in their [docs](https://github.com/styled-components/styled-components#third-party-components),  the only requirement to use styled-components with third-party components is the following:

> Note: styled-components generate a real stylesheet with classes. The class names are then passed to the react component (including third party components) via the className prop. For the styles to be applied, third-party components must attach the passed-in className prop to a DOM node.

You only need to pass `this.props.Classname` to the wrapper div's className prop on `BoostrapTable.js`. 

```javascript
classSet('react-bs-table-container', this.props.className, this.props.containerClass)
```

This is fully compatible with all react-bootstrap-table's customizable classes.

### DIFF
--- a/examples/js/expandRow/demo.js
+++ b/examples/js/expandRow/demo.js
@@ -1,6 +1,6 @@
 /* eslint max-len: 0 */
 import React from 'react';
-import ExpandRow from './ExpandRow';
+import ExpandRow from './expandRow';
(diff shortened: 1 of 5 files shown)


AllenFang:
Merged!  Thanks you :)

PR #968, opened 2017-01-17
Correctly applies border-radius to start and end pagination buttons.

bencooling:
The border-radius css property is correctly applied now by filtering out previously hidden & disabled buttons from being rendered in the pagination navigation. This means that whether the button is "1" or "First" or "Previous" it will correctly display the rounded left corners and whether the button is "10" (for example) or "Last" or "Next" it will correctly display the rounded right corners.


AllenFang:
@bencooling, ok, I'll check this out. Thanks your contributions 👍

bencooling:
I have come across an issue with the logic for determining whether the page being shown is the end, I will push a fix shortly

AllenFang:
Released on `3.0.0-beta.8`, thanks your contribution 👍

PR #993, opened 2017-01-26
Allow percentage values in column width. Fixes #910

### LINKED ISSUE #910: Setting column width with percentage

ptitFicus:
I accidentally deleted the repo of PR #923 sorry.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -64,11 +64,10 @@ export default {
         display: column.hidden ? 'none' : null
       };
       if (column.width) {
-        const width = parseInt(column.width, 10);
-        style.width = width;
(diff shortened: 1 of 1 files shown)


AllenFang:
thanks, because I fixed the column width unalign problems and refactor the strategy of adjusting column dynamically so that I thought configure the column with with percentage is possible, I'll do more test, thank you :0

AllenFang:
Released on `v2.11.0`, Thanks your contribution 👍

Issue #973, opened 2017-01-19
Sort not working, don't quite understand the error message


cheapsteak:
Hi there!

I'm getting this error when I try to sort on a column, not quite sure what it means

```
The type of sort field and order should be both with String or Array
```

I have provided a sortFunc

AllenFang:
@cheapsteak, could you please show me your code and did you use multiple sort?

cheapsteak:
Yup! Here you go

```js
<BootstrapTable
  data={this.reports}
  keyField="id"
>
  <TableHeaderColumn
    dataSort
    dataFormat={(cell, row) => row.blubrry.periodDownloads}
    sortFunc={(a, b, order, sortField, path) => (_.get(a, path) >= _.get(b, path) ? 1 : -1) * (order === 'asc' ? 1 : -1)}
    sortFuncExtraData={'blubbry.periodDownloads'}
  >Blubrry Weekly Downloads</TableHeaderColumn>

</BootstrapTable>
```

AllenFang:
@cheapsteak, I guess because you dont assign a `dataField` on  `TableHeaderColumn`, could you help me to add a `dataField` as a string value on it?

cheapsteak:
That works! I assigned an empty string to `dataField` and sorting is now working. Thanks!

cheapsteak:
Sorry nvm, the sort order seems inconsistent, not sorting as expected :\

AllenFang:
> the sort order seems inconsistent

sorry, is revert? or what?

cheapsteak:
Sorry about that, here's a pen!

http://codepen.io/cheapsteak/pen/MJmbzV

Behaviour is kind of odd, it only sorts after clicking the table header twice, and the sort indicator arrows don't quite match with the sort order

AllenFang:
ok, that's a great example 👍

varak69:
Hi @cheapsteak 

Your codepen isn't sorting because you misspelled 'blubrry' in line 26

cheapsteak:
Oh my god.
So sorry about that X|

varak69:
@cheapsteak No worries, it happens to all of us :P

As for your initial error:
The type of sort field and order should be both with String or Array

If you specify an initial sort order in the options then you have to specify both sortName and sortOrder. If not using multiColumnSort then you can specify them as string:
<BootstrapTable options={{ sortName: 'periodDownloads', sortOrder: 'asc' }}>
...
[...]

PR #996, opened 2017-01-30
Fixed sorting bug for multi column sort

varak69:
There was a bug in multi column sort where the orders and sort fields  would be reversed on table redraws without the user changing the sort info. This was because array.reverse() was used which changes the underlying array. Applying slice().reverse() first makes a copy of the original array and then returns the reverse of it, leaving the original arrays in a correct state.

### DIFF
--- a/src/store/TableDataStore.js
+++ b/src/store/TableDataStore.js
@@ -57,8 +57,8 @@ export class TableDataStore {
       if (order.length !== sortField.length) {
         throw new Error('The length of sort fields and orders should be equivalent');
       }
-      order = order.reverse();
-      this.sortList = sortField.reverse().map((field, i) => {
+      order = order.slice().reverse();
+      this.sortList = sortField.slice().reverse().map((field, i) => {
         return {
           order: order[i],
           sortField: field


AllenFang:
hello @varak69, thank your contribution, I'll review it as soon as possible, btw, is https://github.com/AllenFang/react-bootstrap-table/issues/973#issuecomment-274288558 the issue you try to solve?

varak69:
Hi @AllenFang 

I took a look at that issue and commented as to why it isn't working. Just a spelling error.

This fix is for another issue which hasn't been logged but I came across it in our app. We had a table which was being unnecessarily redrawn on mouse enter/leave. Each time it was redrawn the order of the sort fields changed without them being changed by the user. I'll see if I can make a codepen for it.

AllenFang:
thanks for quickly response :) it's awesome that you found the root cause for that issue, just a wrong spell.. 👍

varak69:
@AllenFang 

Here's a codepen:
http://codepen.io/Varak/pen/egVzWz?editors=0010

As you can see, the sort order isn't changed anywhere in the codepen yet when the component rerenders (just move the mouse over it multiple times), the sort field and sort order array are reversed each time.

AllenFang:
Released on `v2.11.0`, thanks your contribution !!

varak69:
Thank you! :)

PR #998, opened 2017-01-30
TableFilter - custom palceholders, className

rolandsusans:
Now its possible to 
1) set custom placeholder for TableFilter component (current implementation will not be affected, because if filterPalceholder is not defined, previous name will be used - as its currently implemented) 
image
2) It's possible to pass custom class name for TableFilter, currently implementation should not be affected


AllenFang:
@rolandsusans, thanks your contributions 👍  I'll review and merge it as soon as possible.

AllenFang:
@rolandsusans, after I reviewed this PR, I saw you try to improve(add class & placeholder) the filter on the footer, actually, `react-bootstrap-table` have two kind of filter, one is this, another is in the header.  
You can check [these](https://github.com/AllenFang/react-bootstrap-table/tree/master/examples/js/column-filter) example, let me call it as **column filter on header** and most of people use it, it has multiple filter type(text, select, date etc.) and have more ability to custom it, like placeholder, filter format, options etc.

Actually, In the `v3.0.0-beta`, I wanna this both filter to be the same but just in the different place and all the filter have the same ability and functionality from  **column filter on header**.

and the placeholder is duplicated function on  **column filter on header**, but the class name is a good idea that I haven't this functionality, so I gonna close this PR and I'll start to do this job on `v3.0.0-beta` in near future. Sorry for wasting your time.

Let me know if you have any concern.

rolandsusans:
@AllenFang i'm thinking about search filter in column headers 
image

Agree, that current implementation of "column filter on header" is much more flexible. It will be nice feature to have some kind of prop for specifying placement for search filters (in column header | under table | or both ). :+1: 

Can't  wait for v3.0.0 release.

PR #1011, opened 2017-02-05
Fix propTypes for tdStyle and thStyle

nickryall:
These were set to Proptypes.Object which results in the propType being  'undefined' .


AllenFang:
@nickryall, thanks for fixing this stupid mistake... thank.

Issue #876, opened 2016-12-08
custom cell edit mode can we have on keyboard events like pressing 'tab' key 'enter' key 


hctslm:
I am having a requirement of table inline editing, I need to allow the user to edit row wise from one cell to other by pressing tab (and the previous cell content should be saved when user moved from one cell to another cell by pressing tab)

Is there any solution to add multiple event to 
const cellEditProp = {
		  mode: ['click','keydown','keypress']
		};
If we achieve this it will a big help to user without using mouse he can edit the table with in less time

Thanks in advance

AllenFang:
@hctslm, thanks for advice, I know this feature will be useful, but currently I have no much time to do this, so PR is welcome :)

If I have time in the future, I'll back to support this feature 👍 
Thanks

samfisher908:
Is anyone working on this right now? I'm planning on giving it a try for a project I'm working on right now and am happy to contribute my work to the codebase if everything goes smoothly.

AllenFang:
@samfisher908, I think there's no one work on it and this issue may start to handle after v3.0.0 release for me(current still on beta), so you can take a look for codebase and let me know if you have any question.

PR is welcome, thank you :)

madeinfree:
Sorry @samfisher908, I didn't see your comment on this work.., #1015 is my solution to resolve this features, if you would like to PR, you just sent your PR welcome :D

madeinfree:
duplicate #693 ?

AllenFang:
Forgot to close it and already fixed on `v3.0.0`

PR #1015, opened 2017-02-07
feature - use tab to change edit row

madeinfree:
Features for #876, try /examples/cell-edit 👍 

tab-to-cell-edit

### DIFF
--- a/src/TableBody.js
+++ b/src/TableBody.js
@@ -65,6 +65,7 @@ class TableBody extends Component {
                 format={ column.format ? format : false }
                 key={ i }
                 blurToSave={ cellEdit.blurToSave }
+                onTab={ this.handleEditCell }
                 rowIndex={ r }
                 colIndex={ i }
                 row={ data }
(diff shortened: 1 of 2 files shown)


AllenFang:
@madeinfree, That's pretty awesome, maybe I can base this PR to support this table with a excel-like moving with table also. Thanks your contributions 👍

madeinfree:
Good works 👍

AllenFang:
Released, https://github.com/AllenFang/react-bootstrap-table/releases/tag/v3.0.0-beta.11

Thanks your contributions

dekelb:
If already supported I think it will be nice to also have the shift+tab to go back to the previous tab. Since the general functionality is already there I think it would be very easy to implement.

madeinfree:
@dekelb Yes, thanks for your remind, I can help this if need it.

dekelb:
It's not something that I need, but I really think it will make this feature complete.

AllenFang:
I'll consider to add this in the future. Thank

PR #1082, opened 2017-02-24
fix an exception if table source data is immutable type

dp-huang:
If table source data is built by immutable object list, e.g. by seamless-immutable or immutable-js library, the table will be broken if list is empty, which is caused by a push action to the list.

issue

### DIFF
--- a/src/TableBody.js
+++ b/src/TableBody.js
@@ -169,15 +169,19 @@ class TableBody extends Component {
       }
       return (result);
     }, this);
+
+    var tableRowsOutput = tableRows;
+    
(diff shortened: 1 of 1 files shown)


AllenFang:
is that cause by the immutable.js can't call the map method on any array?

dp-huang:
Immutable list can call map method, but can't call push method on the list. Above PR avoids such case.

AllenFang:
ok, thanks, I'll merge after my travel.

AllenFang:
Released on `v3.1.1`, thanks your contributions

PR #1155, opened 2017-03-17
	Extending onSortChange to return freshly sorted table data

lequesne:
Wanted to retrieve the freshly sorted table data as a sorted array on return of onSortChange event function.

Have added the sort result as a fourth param to the onSortChange function.


AllenFang:
@lequesne, in my consideration, `onSortChange` is supposed not to return the result, how about I add another hook, like `afterSortChange`? is that help you?

Thanks, and sorry for lately reply, I'm busy at this month.

nirmalasubu:
@AllenFang I have react bootstrap table with pagination. when I click on the sort button on the 3rd page data is sorted but it is not taking back to the first page . I'm expecting sort behavior like this bootstrap table https://datatables.net/examples/styling/bootstrap.html .  Is this afterSortChange  function takes me to the first page whenever I do sorting.

AllenFang:
@nirmalasubu, not a good solution, but you can give a try

```js
class DefaultPaginationTable extends React.Component {
  constructor(props) {
    super(props);
    this.onSortChange = this.onSortChange.bind(this);
  }

  onSortChange() {
    const sizePerPage = this.refs.table.state.sizePerPage;
    this.refs.table.handlePaginationData(1, sizePerPage);
  }

  render() {
    const options = {
      onSortChange: this.onSortChange
    };
    return (
      <div>
        <BootstrapTable
          ref='table'
          data={ products }
          options={ options }
          pagination>
          <TableHeaderColumn dataField='name' dataSort>Product Name</TableHeaderColumn>
          <TableHeaderColumn dataField='price' dataSort>Product Price</TableHeaderColumn>
        </BootstrapTable>
      </div>
    );
  }
}
```

nirmalasubu:
@AllenFang It worked .  Thank you

PR #1178, opened 2017-03-28
Drag & Drop

### LINKED ISSUE #917: Can I make table rows draggable using react-dnd and current APIs?
I find that react-dnd can inject into components and easily make them draggable.
https://github.com/gaearon/react-dnd
Is it possible to customize my own draggable table using it?
Do we expose any API to wrap the table rows?

benlangfeld:
This implements drag&drop support and as such closes #917. I would of course squash this to a single commit for it to be merged, but thought it potentially useful to retain the history for an initial review.


AllenFang:
@benlangfeld, actually, it really pretty awesome and cool thing for `react-bootstrap-table`. But there're two thing is what i concern:

1. Too many new dependencies are added into this repo, most of table or grid react component which have clean dependency, because it just only table. Actually, `react-bootstrap-table` depend on `react-modal` and `react-toastr` is a big mistake for me, there're a lot of issues about them. Finally I figure out  these kind of feature maybe use plugin or something like that to implement(But I know it's hard and my code base is not good actually)...(It' should not embed in the table component directly)

2. I'm planning the next major phase, that I want to fix the unalign issue and make table responsive and become single table instead of two table structure, so there will be a big change in the future.

Anyway, I really sorry to say that I may not merge your PR, but it's doesn't mean you had a bad feature or something, it's just my feature plan. Sorry about that, but you still did a great job, it's a awesome feature, thanks your hard work

benlangfeld:
Thanks for your feedback. We'll proceed to publish this as a fork, since this module doesn't have any plugin infrastructure we can use to inject this.

AllenFang:
> since this module doesn't have any plugin infrastructure we can use to inject this.

yes, it's also the defect from the beginning design

Issue #1154, opened 2017-03-16
Making cell readonly


ibilanchuk:
Hello, I want to make some cells non editable. Not fully row, only one cell. Is there any solution this problem?

bennnjamin:
Yes, set editable to false like this:

`<TableHeaderColumn dataField="name" editable={false} dataSort {true}>Name</TableHeaderColumn>`

AllenFang:
@bennnjamin, great job 👍 

Let me know if you have any questions, thanks

ibilanchuk:
@bennnjamin, your solution block editing on column. But i need block cell. For example cell is situated in [1,2] - first row,second column

bennnjamin:
@ibilanchuk There is no way built in. The closest you can get is to implement `beforeSaveCell` http://allenfang.github.io/react-bootstrap-table/docs.html#beforeSaveCell and `return false` for the cells you do not want to save edits for. Otherwise you will have to implement this yourself.

AllenFang:
sorry for lately, reply, actually, the `editable` can accept a bool value or a function, if you use the function, there're four arguments, `cellValue`, `row`, `rowIndex` and `columnIndex`. in addition, this function is supposed to be return a bool value to tell `react-bootstrap-table` this cell is editable or not.

Let me know if you got any questions, thanks

AllenFang:
BTW, I should update the docs, so open issue again.

AllenFang:
Docs updated for `editable`, check http://allenfang.github.io/react-bootstrap-table/docs.html#editable

bennnjamin:
@AllenFang I discovered another issue with this. If you disable editing for the cell, there is no way to apply validation onAddRow. For example, what if you want to insert a new row into the table but after that you want the cell to be read only? You need validation on insert, but not the ability to edit the cell.

PR #1179, opened 2017-03-28
Make Cell Read Only

bennnjamin:
This is to implement a new option on the `editable` property for Table Columns as discussed in Issue #1154. 

`editable` now accepts a property `readOnly` which when set to `true` will make the column read only (i.e. disable cell editing). The current implementation of setting `editable` to `false` will prevent **disable column input for adding a new row**. 

This is useful if you want to implement `INSERT` but not `UPDATE` on your Table.


AllenFang:
I'll spend some time to review your PR

bennnjamin:
The dist folder came from cloning the master branch. I have since updated the fork to not include those folders

bennnjamin:
I added the`dist` folder back, please check now.

AllenFang:
Released on `v3.1.6`, thanks your contributions :)

bennnjamin:
Thanks for merging this!

PR #1190, opened 2017-04-01
Added has footer option

myapos:
- Added hasFooter attribute
- Useful in reporting total calculations
- Presents footer row always in bottom of the table while the rest of the data in table is sorted normally
- Added an example in examples folder custom-sort-table-has-footer.js


myapos:
Hi @AllenFang. hasFooter option gives the possibility to the user to display total data such as sums always in the last row exactly as a footer row. 
The problem was that the last row was sorted among the other rows.You can see this in the example that i added in the corresponding folder (examples/js/sort/custom-sort-table-has-footer.js) I added the last row with the sums after the generation of the data products.  So in this case i extract the last element and get all the other elements except the last one in a new subArr array, (see if you want the function sort in TableDataStore.js. After that i apply sorting in the subAr and rejoin the sorted subAr (0, length -1) with the last element. I did the fetch upstream again as you can see. Thank you for reviewing this!

AllenFang:
ok, I'll review it, thanks your explanation 👍

anikakalbrun:
Hello @AllenFang. Are you planning to add total calculation feature in footer as above in upcoming release? I found that feature really needed. Please let me know your thoughts. Thank you for great plugin!

AllenFang:
@anikakalbrun, yes, I will support this but I can't promise when I can done for it, sorry

Skeetfly:
How is this PR going?

AllenFang:
As I mentioned on readme, `react-bootstrap-table` already stop implement any new feature and forward to `react-bootstrap-table2` https://github.com/react-bootstrap-table/react-bootstrap-table2/tree/develop

And the table footer will be fully support in `react-bootstrap-table2`

PR #1196, opened 2017-04-04
Conjunct Multi Column Search

CarstenWickner:
Use Case: the user enters one search term and still sees too many entries: they just have to enter a second search term (separated by a whitespace) to narrow the filter further down.

----

The `multiColumnSearch` prop allows for multiple terms to be entered into a search field in order to display all rows that contain at least one of the given terms.

With the introduction of an additional `strictSearch` prop, we can achieve and AND search filter, i.e. only those rows are displayed that contain all of the search terms (in whatever order). In combination with the `multiColumnSearch` this allows for such an AND search filter over all `searchable` columns.

If unspecified, the `strictSearch` will be set to `!multiColumnSearch` to preserve the current default behaviour.

### DIFF
--- a/src/store/TableDataStore.js
+++ b/src/store/TableDataStore.js
@@ -18,6 +18,7 @@ export class TableDataStore {
     this.pageObj = {};
     this.selected = [];
     this.multiColumnSearch = false;
+    this.strictSearch = true;
     this.multiColumnSort = 1;
     this.showOnlySelected = false;
     this.remote = false; // remote data
@@ -29,6 +30,7 @@ export class TableDataStore {
     this.colInfos = props.colInfos;
     this.remote = props.remote;
     this.multiColumnSearch = props.multiColumnSearch;
+    this.strictSearch = props.strictSearch;
     this.multiColumnSort = props.multiColumnSort;
   }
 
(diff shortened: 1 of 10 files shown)


src/store/TableDataStore.js
```diff
@@ -18,6 +18,7 @@ export class TableDataStore {
     this.pageObj = {};
     this.selected = [];
     this.multiColumnSearch = false;
+    this.strictSearch = true;
```
CarstenWickner: According to that reasoning we can also remove the `this.multiColumnSearch = false` line above.
I just wanted to be consistent 😉

CarstenWickner:
I added some basic tests to cover the comparison table on issue #1199.
However, the travis setup expects a too old node version and therefore fails the automated test runs.

CarstenWickner:
@AllenFang is there anything else you want me to change? Otherwise i'd assume this is good to be merged now. 😁

AllenFang:
Thanks for your hard work, I'll start to review you PR 💯

AllenFang:
@CarstenWickner, I'm going to merge and release in this weekend, so if you have other thing not to do yet, open an another PR is welcome 👍

AllenFang:
Release at `v3.2.0`, thanks your hard work 👍

PR #1198, opened 2017-04-04
Fixing problem with externally controlling expanded rows.

dangerbell:
Changing the "expanding" option was not changing which rows were selected. Setting an initial list of expanded rows worked, but modifying that list through the props to BootstrapTable did not work.

The value being passed down to TableBody is `this.state.expanding`, but that is only set from props as part of the constructor, so changes to props.options.expanding of an already mounted BootstrapTable weren't being taken into account.

I've added some handling in componentWillReceiveProps to deal with change to `options.expanding`. I've copied the pattern used for the pagination options.

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -240,6 +240,14 @@ class BootstrapTable extends Component {
       });
     }
 
+    // If setting the expanded rows is being handled externally
+    // then overwrite the current expanded rows.
+    if (this.props.options.expanding !== options.expanding) {
+      this.setState({
+        expanding: options.expanding || []
+      });
+    }
+
     if (selectRow && selectRow.selected) {
       // set default select rows to store.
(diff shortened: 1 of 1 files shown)


dangerbell:
Looks like this build is failing because AllenFang/react-toastr doesn't exist. I assume that's a problem with where I branched off of. Can you tell me what's needed to fix that?

AllenFang:
Thanks your contribution, I'll test it again 👍

dangerbell:
Thanks! I didn't run `gulp prod` on it, should I have done that?

AllenFang:
> Thanks! I didn't run gulp prod on it, should I have done that?

ah, that's my job :) it's for building production before release :) Thanks 👍

AllenFang:
Released on `v3.1.7` and thanks your contributions :)

PR #1203, opened 2017-04-06
Sticky Headers

walkman2058:
Hello, I added a functionality for having sticky headers. To show it there needs to be set a stickyHeaders={ true } as props.
I also added an example you can find it here react-bootstrap-table\examples\js\others\table-with-sticky-headers.js

### DIFF
--- a/examples/js/others/demo.js
+++ b/examples/js/others/demo.js
@@ -3,6 +3,7 @@ import MouseEventTable from './mouse-event-table';
 import TableInTabs from './table-in-tabs';
 import GetPageNumByKeyTable from './expose-api-table';
 import PrintableTable from './printable-table';
+import TableWithStickyHeaders from './table-with-sticky-headers';
 import { Col, Panel } from 'react-bootstrap';
 
 class Demo extends React.Component {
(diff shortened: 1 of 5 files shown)


AllenFang:
@walkman2058, it's a cool feature, but did we have a chance that dont use the `react-overlays`? Thanks 👍

vKyrylenko:
I think that it is possible, but it will require additional functionality/component with the same functionality for calculation position, affix, offsetTop, offsetBottom as in react-overlays.
Why are you against using react-overlays ?

AllenFang:
@vKyrylenko, because I dont want to add too many dependency into this lib. I doesn't look into you code to see how it work, but if we can make this cool feature plugable or use it as decorator or HOC will be good.

My final goal is if the user want to sticky-headers, the can use/install `react-overlays` in their repo and use a HOC or decorator with `react-bootstrap-table` to have this kind of feature, so that means, native `react-bootstrap-table` doesn't support sticky-headers and no dependency with `react-overlays `.
That's what I think, and It's a dream and most likely is hard to achieve, because I dont have a good components design now and actually I don't start to think how to implement it, it just a idea.

I'll spend sometime to toward this goal for your PR in these days and let me know if you have any idea.

FloMaetschke:
What happened to this now?

FloMaetschke:
It would be nice, to be able to specify the docking top height, e.g. if there is a navbar docked already to place the thead below it.

AllenFang:
@FloMaetschke `react-bootstrap-table` already stop implement new feature and move to [react-bootstrap-table2](https://github.com/react-bootstrap-table/react-bootstrap-table2/tree/develop). I will  consider to implement sticky header seriously.

PR #1215, opened 2017-04-11
clean custom delete btn example

uralbash:


AllenFang:
Thanks 👍  I think there're still have the same minor issue in this repo ;p

Issue #1187, opened 2017-03-31
Remote mode : Problem with OnFilter change and OnPageChange


remy-poirier:
Hello,

I will try to be as clear as possible.

I use your library with the remote mode, I basically do filtering, sortering, and pagination. 
I recently implemented the filter by column, and implemented the onFilterChange.

It works pretty well if all I do is filtering, but when I try to change the page with this filter, my handler to change page is called, and the data is refreshed, so that's ok for this, it's the good data that is displayed, but the problem is that the page highlighted is still the page 1 whereas the value of the currentPage in the state is 2 and this is this value that is passed to the `page` parameter of my `options` object that I pass to my Table.

Do you have any idea ?

AllenFang:
@remy-poirier, could you have a simple and minimal for me to reproduce this issue? thanks 👍

AllenFang:
BTW, is that same condition with https://github.com/AllenFang/react-bootstrap-table/issues/1037 ?

AllenFang:
Fixed on `v3.1.8`

BenoitAverty:
I still have this problem in version 3.2.0. I thinks it's linked to the #1037 issue yes.

What I observed is this : When a filter is active, and the page is changed, it fires the `onFilterChange` callback. This makes the `state.currPage` reset to 1 (as discussed in #1037) even though the `props.options.page` does not change. This causes a mismatch between `props.options.page` and `state.currPage`.

It boils down to the same think I said in issue #1037 : when the table is in remote mode, the `state.currPage` should **always** be equal to` options.page`.

I know you don't agree, so as a workaround I made the page in my component be restet to one when filter changes. But in that case, I can't do it : that would make the page impossible to change when a filter is active !

This doesn't seem to happen for text filters, because the filterchange is not called when the page changes.

BenoitAverty:
@remy-poirier Can you tell us if you still have the issue with the latest version ?

BenoitAverty:
One solution would be to change the condition at https://github.com/AllenFang/react-bootstrap-table/blob/master/src/filters/Select.js#L27 so that the filter is not fired when changing the options object. By making a deep comparison instead of object comparison this would not fire the event unless necessary

AllenFang:
@BenoitAverty, this is is not related with #1037

For your PR, is for https://github.com/AllenFang/react-bootstrap-table/issues/1174 also?

Thanks 👍

BenoitAverty:
[...]

Issue #1174, opened 2017-03-26
Enhance componentDidUpdate for select filter


AllenFang:
https://github.com/AllenFang/react-bootstrap-table/commit/d3434f2feba8ac5182dc082a29a247c5228d996f#commitcomment-21471262

PR #1260, opened 2017-04-26
Fix callback condition in Select filter

BenoitAverty:
The filter changed callback will be called less often, and in particular it will not be called when the options object has changed but has the same entries. This should fix issue #1187 

This is only a proposition, and other strategies are possible. In particular, I implemented the comparison function because the options object is always flat. It is also possible to use an external package like deep-equals or similar.

### DIFF
--- a/src/filters/Select.js
+++ b/src/filters/Select.js
@@ -21,10 +21,21 @@ class SelectFilter extends Component {
   }
 
   componentDidUpdate(prevProps) {
+    function optionsEquals(options1, options2) {
+      const keys = Object.keys(options1);
+      for(let k in keys) {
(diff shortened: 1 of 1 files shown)


AllenFang:
is try to fix https://github.com/AllenFang/react-bootstrap-table/issues/1174?

BenoitAverty:
Yes this would also fix #1174

AllenFang:
@BenoitAverty, thanks 👍  I'll merge it ASAP.

AllenFang:
Released on `v3.3.0`, thanks your contribution.

BenoitAverty:
You're welcome, it's the least I can do :)

PR #1262, opened 2017-04-26
Restrict border & border-radius to bordered tables

sman591:
As seen on the [Bootstrap table docs](http://getbootstrap.com/css/#tables), the default ("basic") style is a table without any borders, other than the border separating rows.

Therefore, tables generated with `react-boostrap-table` with borders disabled shouldn't have a border, either.

I didn't create an example for this (happy to if wanted), but this is what it looks like:

```jsx
<BootstrapTable data={ products } bordered={ false }>
...
```

screenshot 2017-04-26 16 56 17

### DIFF
--- a/css/react-bootstrap-table.css
+++ b/css/react-bootstrap-table.css
@@ -5,7 +5,7 @@
   margin-bottom: 0;
 }
 
-.react-bs-table {
+.react-bs-table-bordered {
   border: 1px solid #ddd;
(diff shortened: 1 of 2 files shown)


AllenFang (approved):
LGTM

AllenFang:

MaheshK02:
header border i need to make it 1px solid #ddd;

unable to change

Issue #414, opened 2016-04-20
Remove margin on react-bs-table


johannessjoberg:
Hi,

I have a problem with the margin on react-bs-table class. I can't find any attribute in the docs to alter it. It would be nice to be able to alter the styling on the main component(BootstrapTable) without having to override that specific class.

Thanks for an awesome repo!
Cheers Johannes

AllenFang:
Thanks @johannessjoberg, there's no way to define some style on ```react-bs-table``` through react(Maybe I can support it), but you can override the styling by your css.  Feel free to discuss with me :)

johannessjoberg:
All right, good to know :) 

I'm going pretty heavy on the inline-styling with this project, i.e. not using classNames, and want to avoid external css as much as possible(death to globals ;) ).

Maybe the possibility to be able to set a `style` attribute on the BootstrapTable component which would only effect the styling on the wrapping div would be a solution?

AllenFang:
> I'm going pretty heavy on the inline-styling with this project, i.e. not using classNames, and want to avoid external css as much as possible(death to globals ;)

that's a right thing on development a react application.



> Maybe the possibility to be able to set a style attribute on the BootstrapTable component which would only effect the styling on the wrapping div would be a solution?

I'll try to make a custom inline-styling on ```react-bs-table``` and apply a props on ```BootstrapTable```

johannessjoberg:
Sweet! 
Perhaps it would be nice to implement it on other components like `TableHeaderColumn` as well?

Btw, thanks for the fast response!

AllenFang:
Nice suggestion, btw, TableHeaderColumn apply ```classname``` for user to customize there column, [there](https://github.com/AllenFang/react-bootstrap-table/tree/master/examples/js/style) are some useful example.  But it's not a inline-style way.

AllenFang:
@johannessjoberg, you can check [this](https://github.com/AllenFang/react-bootstrap-table/blob/master/examples/js/style/inline-style-table.js) example to see how to apply inline styling. I support four type of inline style(```containerStyle```, ```tableStyle```, ```headerStyle```, ```bodyStyle```), it's corresponding to the class of  ```react-bs-table-container```, ```react-bs-table```, ```react-bs-container-header``` and ```react-bs-container-body```.

johannessjoberg:
@AllenFang Nice! Thank you for the fast response and also the quick implementation. Awesome job, awesome repository!

PR #1263, opened 2017-04-26
Maintain default bootstrap table margins

sman591:
Issue #414 touches on how this library adds margins surrounding the table, which was addressed in e60e607ca8a01ec89cd728196f083636bd7f99bf by allowing a `style="..."` override. However, I don't feel this is the right approach to this specific issue.

Bootstrap tables don’t have any surrounding margins by default*. Therefore when using this library to create bootstrap tables, there shouldn’t be any surrounding margins by default, either.

This PR maintains the intended margins between components inside react-bootstrap-table (such as the toolbar or pagination area), while removing the margin surrounding the table.

\* Bootstrap tables do have a margin-bottom of 20px that their docs override, however I don’t think this PR is the right place to address that.

Original "complex" example:
screenshot 2017-04-26 17 30 31

"Complex" example after these changes:
screenshot 2017-04-26 17 30 10

### DIFF
--- a/css/react-bootstrap-table.css
+++ b/css/react-bootstrap-table.css
@@ -22,16 +22,12 @@
   text-overflow: ellipsis;
 }
 
-.react-bs-table {
-  margin: 5px 10px 5px 10px;
-}
-
 .react-bs-table-pagination {
-  margin: 10px;
+  margin-top: 10px;
 }
 
 .react-bs-table-tool-bar {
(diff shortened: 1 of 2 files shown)


AllenFang:
@sman591, I'll check this out, thanks your contribution 👍

sman591:
@AllenFang no problem, thanks for the great library! Let me know if you have any questions

AllenFang:

PR #1264, opened 2017-04-27
Enable custom sort (issue #1256)

### LINKED ISSUE #1256: Server side sort
Hi! Great table component!
I'm using react-bootstrap-table as a view table only which triggers sorting on the server,
that's why I need the sort indicators on the column headers to display correct data. In other words I want to disable the the native component sort but to still enjoy the directional indicators.

I'm loading the table like so:
`TableWrapper.jsx`
```
    remote(remoteObj){
       return{
         ...remoteObj,
         sort: true,
       }
    }

   onSortChange(key,order){
      browserHistory.getCurrentLocation();
      browserHistory.push({
         ...location,
          query: {
           ...location.query,
           sortField: sortKey,
           sortDirection: order
         }
      });
   }

    const tableOptions = {
      sortName: this.props.sortField,
      sortOrder: this.props.sortDirection,
      onSortChange: this.onSortChange.bind(this),
    };

    return (
      <BootstrapTable remote={this.remote} data={this.props.items} options={tableOptions}>
        {this.props.children}
      </BootstrapTable>
    )
```

When the `onSortChange` is fired I push to history and the data is refetched (already sorted nicely by the server)..
Problem is, when one sets the `sortName` & `sortOrder` props, bootstrap table automatically tries to sort it again: 
These are the lines causing the sort inside `BootstrapTable.js` (more context [here](https://github.com/AllenFang/react-bootstrap-table/blob/30cad00731488a2733a244e24ea861e4f7e47e00/src/BootstrapTable.js#L112-L115)) 
 
```
    if (sortName && sortOrder) {
      this.store.setSortInfo(sortOrder, sortName);
      this.store.sort();
    }
```
IMHO if sortName and sortOrder are present it's still should not sort the data without checking the `remote` first (if I understood `remote` correctly). That's why I'm suggesting this edit:
```
      if (sortName && sortOrder) {
        this.store.setSortInfo(sortOrder, sortName);
        if(this.props.remote().sort !== true){
          this.store.sort();
        } 
      }
```
I'll be happy to submit a PR for this if this is the right way to go.

Paveltarno:
resolve #1256 
Iv'e only changed the src so it can all be compiled later upon release.
LMK if there is anything else [=

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -109,11 +109,14 @@ class BootstrapTable extends Component {
     const sortName = options.defaultSortName || options.sortName;
     const sortOrder = options.defaultSortOrder || options.sortOrder;
     const searchText = options.defaultSearch;
+
     if (sortName && sortOrder) {
       this.store.setSortInfo(sortOrder, sortName);
-      this.store.sort();
+      if (this.props.remote().sort !== true) {
+        this.store.sort();
+      }
     }
-
+    
     if (searchText) {
       this.store.search(searchText);
     }


AllenFang:
thanks, I'll review it 👍

AllenFang:

PR #1268, opened 2017-04-27
Fix bug when default DateFilter and date editor are used together

JSterling8:
When the DateFilter is used in conjunction with the builtin datetime field editor, TableDataStore.js throws an error. TableDataStore expects a Date, but the field editor (basically `<input type='datetime' />` returns a string. This needs to be checked for in TableDataStore.js, and the variable converted into a Date object if it's not already.

Along the same lines, the dateFormatter(cell, row) function in the examples also needs a similar check in order for both features (the date filter and date editor) to work together.

### DIFF
--- a/examples/js/column-filter/all-filters.js
+++ b/examples/js/column-filter/all-filters.js
@@ -36,6 +36,10 @@ function enumFormatter(cell, row, enumObject) {
 }
 
 function dateFormatter(cell, row) {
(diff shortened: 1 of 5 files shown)


AllenFang:
@JSterling8, I'll check this our on next week, if any question will ask you, but thanks your contributions.

AllenFang:
Released on `v3.3.7`, thanks your contribution

PR #802, opened 2016-10-29
use yarn, upgrade react-router v3, fixed issue

madeinfree:
Hi @AllenFang, 
# Yarn install

This is PR with yarn and dependencies fix
## Install yarn

``` command
> npm install yarn --global
> yarn
```
# Fixed dependencies issue
## React, React Bootstrap

use react version 15.3.0 cause PropTypes[#7132](https://github.com/facebook/react/pull/7132) issue from react-bootstrap v0.29.5 so I upgrade react-bootstrap v0.30.0[#2095](https://github.com/react-bootstrap/react-bootstrap/pull/2095) to resolve this issue.
## React Router

I try to upgrade react-router to avoid history dependencies issue by version 3.0.0([link](https://github.com/ReactTraining/react-router/blob/v3.0.0/package.json#L35)) which use `history` dependencies itself, so we can didn't care about `history` dependencies.

use 

``` javascript
import { hashHistory } from 'react-router' 
//and add props into
<Router history={ hashHistory }>
```

now react-router 3.0.0 working fine.
# Now dependencies version
1. react-bootstrap v3.0.6
2. react-router v3.0.0
# Test working version

react, react-dom: 0.14.7, 15.0.0, 15.3.0 (all pass)

Have some suggestion feel free to discussion with me 👍

PR #1279, opened 2017-05-03
Update Yarn lockfile

sman591:
Looks like Yarn support was added in #802, however the lockfile is out of date with what package.json defines.

Simply pulling down the master branch and running `yarn` produces these changes.


AllenFang:
@sman591, thanks :)

PR #1291, opened 2017-05-07
Pagination dropdown callbacks

### LINKED ISSUE #1153: Pagination dropdown getting close when setState({ data })
I'm using react-bootstrap-table with remote data and fetching the data every few seconds to update the table.

    <BootstrapTable
        remote={ true }
        data={ this.state.data }
        ....

And when fetching the data I use `setState({ data })`.

If I open the pagination menu - once the component is refreshed with the data from the server - the dropdown is getting closed automatically.

I'm using version 3.0.0

Is there something I can do to keep the pagination dropdown open?

dekelb:
This PR will add callbacks to the pagination dropdowns.
This will fix #1153 

Also note that I added the jquery as a plugin for webpack (and import jquery and bootstrap js files inside the bootstrap table). This way we don't need to use the `<script>` files in the html file and everything is within the package.

Thanks!


dekelb:
@AllenFang can you please explain why? There are many questions/issues opened regarding dropdown doesn't work because the need to include the jquery/bootstrap explicitly. If we add this directly inside the package it will solve all of these issues (and we already have this in the package dependencies).

dekelb:
@AllenFang you said that after v3.0.0 you removed jquery and bootstrap but it's not correct. For the pagination dropdown to work you must have both jquery and bootstrap.

If we remove both of them it's great, but currently we need them for the package to work.

AllenFang:
> an you please explain why? There are many questions/issues opened regarding dropdown doesn't work because the need to include the jquery/bootstrap explicitly.

I control the dropdown open/close via React state instead of `jq` and `bootstrap.js`.
<hr/>
Before `v3.0.0`, many people said dropdown is not work, due to they forgot to include the `jquery` and `bootstrap.js` and **I dont include this both dependency explicitly in this project**.  But now it work well.
<hr/>
Fix this problem, you can have two options

1. add `jquery` and `bootstrap.js` explicitly
2. controll by react

I just options two, due options one have following cons:

1. make this lib bigger 
2. it only solve dropdown, not very worth it
3. the user may load duplicate `bootstrap.js` or `jquery`

my question is why you still want to choose option one even if there're some drawback??

> it's not correct. For the pagination dropdown to work you must have both jquery and bootstrap.

you mean there still have bug that dropdown broken? I think it's impossible in `v3.0.0`

robbieyng:
I'm also experiencing issues with the dropdown. I have to explicitly reqiure jQuery and Bootstrap.js for it to work properly.

AllenFang:
@robbieyng, impossible, in `v3.0.0` shouldn't need jquery and bootstrap.js anymore, check your version.

dekelb:
@AllenFang sorry for the late response, it took some time to get back to this pr.
Regarding your comment - this is not correct.
Here is an example of ver3.3.8 without jquery and without bootstrap.js: http://jsfiddle.net/r3poLmmq/
As you can see - the dropdown can *only open* but you can't close it (only if you click on the dropdown itself, not if you click outside the dropdown).
Here is gif showing the problem:
react-bt-dropdown-not-closing

Here is exactly the same code, but with jquery & bootstrap.js: http://jsfiddle.net/3rj65dt1/
Here is a gif showing what the dropdown close when you click outside of it:
react-bt-dropdown-closing
(Note that there is still some problem that you need to double-click on the dropdown to open it, but it's another issue).

PR #1330, opened 2017-05-18
remove warnings caused by PropTypes

SidKwok:
import PropTypes directly from react may cause the warning in browser like this:
```
Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.
```

So I use the `prop-types` library to remove warnings. Noted that tests are all passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   },
   "dependencies": {
     "classnames": "^2.1.2",
+    "prop-types": "^15.5.10",
     "react-modal": "^1.4.0",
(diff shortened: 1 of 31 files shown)


AllenFang:
@SidKwok, thanks for your contribution 👍 but could you please help me to change the target branch to `v4.0.0-dev`?

kenny26:
@AllenFang I already change it to version 4.0.0-beta.2, but the warning still showed up

AllenFang:
@kenny26, this issue not fix yet

Issue #1161, opened 2017-03-20
Column Width does not work when header grouping


ThomasVuillaume:
Hello everyone,

Just encountered a bug, when trying to assign some width to the last TableHeaderColumn of my table.

My code:

```javascript
const daysCalendarColumns = [];
    this.state.colMap.forEach(column => {
      daysCalendarColumns.push(<TableHeaderColumn columnClassName="headerDayNumCol" key={'up_' + column.colDataField} row="0" dataAlign="center">{column.colDayNumber}</TableHeaderColumn>);
      daysCalendarColumns.push(<TableHeaderColumn columnClassName={this.getDayColumnStyle(column.colDayName)} key={'dwn_' + column.colDataField} row="1" customEditor={{getElement: createLoadEditor, customEditorParameters: {dataField: column.colDataField}}} dataField={column.colDataField} dataAlign="center">{column.colDayName}</TableHeaderColumn>);
    });

    return (
      <div style={divStyle}>
        <BootstrapTable data={crasData} cellEdit={cellEditProp} hover tableStyle={{fontSize: 'x-small'}} trClassName={this.getRowClassName}>
          <TableHeaderColumn row="0" rowSpan="2" columnClassName="mainLabelCol" isKey key="mainLabel" dataField="mainLabel" width="140px" editable={false} thStyle={{borderRight: 0}}/>
          <TableHeaderColumn row="0" rowSpan="2" columnClassName="addLabelCol" key="addLabel" dataField="addLabel" width="90px" editable={false} thStyle={{borderLeft: 0}}/>

          {daysCalendarColumns}

          <TableHeaderColumn row="0" rowSpan="2" columnClassName="totalOnRowsCol" key="totalOnRow" dataField="totalLoadOnRow" width="40px" editable={false} dataAlign="right">TOTAL</TableHeaderColumn>
        </BootstrapTable>
      </div>
    );
```

I have some grouped headers standing for day names and nomber of the month. When I try to pu the width on the last column, i got the following error:

> Uncaught TypeError: Cannot read property 'style' of undefined

This comes from webpack:///./~/react-bootstrap-table/lib/BootstrapTable.js @ 1431:

`else {
        _react2.default.Children.forEach(this.props.children, function (child, i) {
          if (child.props.width) {
            header[i].style.width = child.props.width + 'px';
            header[i].style.minWidth = child.props.width + 'px';
          }
        });
      }`

[...]

PR #1363, opened 2017-05-29
fix #1116 - iterate through real children

### LINKED ISSUE #1116: npm run build fails if react-bootstrap-table is enabled
ReactJS
ES6
Webpack
Babel

-- 

In my component I have this : import { BootstrapTable, TableHeaderColumn } from 'react-bootstrap-table';

Then I run "npm run build" >> FAIL

If I comment or remove this line //import { BootstrapTable, TableHeaderColumn } from 'react-bootstrap-table';

Then I run "npm run build" >> OK

Why? And how to solve that? Thanks 

capture d ecran 2017-03-06 a 02 03 31

volodalexey:
Faced with almost the same issue [#1161](https://github.com/AllenFang/react-bootstrap-table/issues/1161) when my header is grouped.
As far as I have understood here we assign row's width to header's width.
Why not just to iterate through real rows instead of children, because some children in this case are just groupped-headers?

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -1271,12 +1271,17 @@ class BootstrapTable extends Component {
         }
       }
     } else {
-      React.Children.forEach(this.props.children.filter(_ => !!_), (child, i) => {
-        if (child.props.width) {
-          header[i].style.width = `${child.props.width}px`;
-          header[i].style.minWidth = `${child.props.width}px`;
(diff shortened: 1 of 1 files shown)


AllenFang:
@volodalexey, thanks your contributions, I'll check this out 👍

AllenFang:
@volodalexey, fixed on `v3.4.0`, thanks your contribution 👍

PR #1388, opened 2017-06-08
Track column index on click events

bleakley:
Click events now track the column index in addition to the row, plus a minor correction to terminology.


AllenFang:
thanks, I'll check this out :)

AllenFang:
@bleakley, thanks your contributions, it was released on `v3.4.1` 👍

PR #1397, opened 2017-06-12
custom CSV separator (unable to use semicolon instead)

Riyouzaki:
Hello guys! I have amounts in my table that uses comma to separate decimal digits, so when exporting my data into CSV  I got some problems since **onExportToCSV** function uses a comma by default. here is a PR that adds  a new prop in options named **exportCSVSeparator**, that's comma by default and can be overridden in BootstrapTable.
I tested It locally and it works fine.
Thank you.

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -984,7 +984,7 @@ class BootstrapTable extends Component {
     } else {
       result = this.store.getDataIgnoringPagination();
     }
-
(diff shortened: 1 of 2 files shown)


AllenFang:
@Riyouzaki thanks your contribution, I'll check it ASAP

AllenFang:
Released on `v3.4.1`, thanks your contribution

Issue #1429, opened 2017-06-28
Issue when running in a headless browser


sgulics:
Hi, 

I am using react-bootstrap-table in a Rails application. The component works fine when running in browsers such Chrome, Firefox, Safari, etc. However, when running in a headless environment such as webkit or phantomjs I get the following error:

` undefined is not an object (evaluating 'child.style.width')`

I traced the issue to this code in [BootstrapTable.js](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/BootstrapTable.js#L1350)

```
for (const i in bodyHeader) {
  if (bodyHeader.hasOwnProperty(i)) {
    const child = bodyHeader[i];
    if (child.style.width) {
      header[i].style.width = child.style.width;
    }
    if (child.style.minWidth) {
      header[i].style.minWidth = child.style.minWidth;
    }
  }
}

```
For some reason, when running in a headless browser, hasOwnProperty  returns true for methods such as 'length', however when running in a normal browser, it will return false for those types of methods. 

If this happens, the 'child' variable becomes an integer, and child.width throws the error. 

My assumption is that it has to do with the polyfills that need to be added to get es5 working in the headless browser. Here is an excerpt from my  webpack config:

```
...
entry: {
    // This will contain the app entry points defined by
    // webpack.client.rails.hot.config and webpack.client.rails.build.config
    'vendor-bundle': [
      'babel-polyfill',
      'es5-shim/es5-shim',
      'es5-shim/es5-sham'

    ],
    'app-bundle': [
      './app/bundles/App/main.js',
    ],
  }
...
```
The issue could be related to my webpack configuration because I am not an expert. However, this issue can easily be resolved by making a small change to the code in BootstrapTable.js,  something like:

```
for (const i in bodyHeader) {
  if (bodyHeader.hasOwnProperty(i)) {
    const child = bodyHeader[i];
    if (child.style) {
      if (child.style.width) {
        header[i].style.width = child.style.width;
      }
      if (child.style.minWidth) {
        header[i].style.minWidth = child.style.minWidth;
      }
    }
  }
}
```
I am using the latest version of babel-polyfill and es5-shim.

PR #1430, opened 2017-06-28
Fix 'undefined is not an object' when running in a headless browser

sgulics:
Here is fix so that react-bootstrap-table can run in a headless browser such as phantomjs or webkit #1429 

I just added a small check to the code in BootstrapTable.js to ensure that the child variable has a style object:

```
for (const i in bodyHeader) {
  if (bodyHeader.hasOwnProperty(i)) {
    const child = bodyHeader[i];
    if (child.style) {
      if (child.style.width) {
        header[i].style.width = child.style.width;
      }
      if (child.style.minWidth) {
        header[i].style.minWidth = child.style.minWidth;
      }
    }
  }
}
```

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -1347,11 +1347,13 @@ class BootstrapTable extends Component {
       for (const i in bodyHeader) {
         if (bodyHeader.hasOwnProperty(i)) {
           const child = bodyHeader[i];
-          if (child.style.width) {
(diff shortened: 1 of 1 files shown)


AllenFang:
thanks @sgulics your contribution, it was released on `v3.4.3`, 👍

sgulics:
@AllenFang no problem. Thank you for creating this project!

PR #1437, opened 2017-07-02
Added an option to exclude the header row from a CSV export.

rhavill:
Thank you for react-bootstrap-table. It is a very useful tool. I would like to add an option, for CSV exports, that allows the header row to be excluded from the file.


AllenFang:
thanks @rhavill your contribution, it was released on `v3.4.3`, 👍

PR #1448, opened 2017-07-04
Feature/revise fix and relative TableHeaderColumn width issue

bbbbighead:
fix table header width adjustment logic when TableHeaderColumn is using fix with or relative width(in percentage %)
- no need to re-calculate and using fix width when scroll is happening, it will cause side effect which will make header width will only grow but not restore to original size when you resize outside window size(drag window large and small back and forth)

- please check the video for how to duplicate the issue
https://youtu.be/-vfrWEJyung

### DIFF
--- a/examples/js/basic/basic-table.js
+++ b/examples/js/basic/basic-table.js
@@ -17,15 +17,15 @@ function addProducts(quantity) {
   }
 }
 
-addProducts(5);
+addProducts(50);
 
 export default class BasicTable extends React.Component {
   render() {
     return (
-      <BootstrapTable data={ products }>
-          <TableHeaderColumn dataField='id' isKey={ true }>Product ID</TableHeaderColumn>
-          <TableHeaderColumn dataField='name'>Product Name</TableHeaderColumn>
-          <TableHeaderColumn dataField='price'>Product Price</TableHeaderColumn>
+      <BootstrapTable data={ products } height='500px'>
+          <TableHeaderColumn dataField='id' isKey={ true } width='100px'>Product ID</TableHeaderColumn>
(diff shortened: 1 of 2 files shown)


AllenFang:
@bbbbighead, you found a known issue, actually, I sacrifice this issues to solve most of unalign issues(although there's still have another unalign issues still there....)

it's a long story that I will not to describe again lol, but we have a lots of unalign issues due to `react-bootstrap-table` use two `div` with `table` to construct a bootstrap table instead of a native `table` tag, so the code you remove is try to solve some unalign issues.

So I can't merge you PR due to it will cause a lots of unalign issue again.... it's very sorry for you...

bbbbighead:
hi @AllenFang , got it! Is there any plan when or which version will un-align issues be fixed? v4.0? or other version.

Or....is there any idea you want to fix this issue? Our team is using your component quite a long time and feel very great about it. Could you could give me some clues and maybe we might provide some helps for this issue.

Thanks for the reply and the effort you made for react community. 
Taiwan No.1 !!!

AllenFang:
Thanks you @bbbbighead :taiwan: 

actually, I usually cant reproduce the unalign issues... and for me, if we want to fix unalign issue thoroughly, the only solution is use one `table` instead of two `div` with `table`, but in single table structure, **it's hard to achieve the horizontal and vertical scroll**, particular at the same time.

I'm not even think that I can fix it in the future  lol
I've plan to rebuilt `react-bootstrap-table` as `react-bootstrap-table2`, which is a new package for replacing `react-bootstrap-table` but I think I don't have too much time, and it can't be done only by myself.

Anyway,  how about I provide a props for you to let you can control if enable/disable width adjusting?
Let me know your concern or question, thanks 👍

PR #1457, opened 2017-07-07
Fixed one small issue for BootstrapTable.js

athenaxing:
int componentWillReceiveProps functions

Add checking code
   if (nextProps.data === undefined) {
      return;
    }
Before 
   this.store.setData(nextProps.data.slice());

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -215,6 +215,9 @@ class BootstrapTable extends Component {
     let { replace } = nextProps;
     replace = replace || this.props.replace;
 
+    if (nextProps.data === undefined) {
+      return;
+    }
     this.store.setData(nextProps.data.slice());
 
     if (!replace) {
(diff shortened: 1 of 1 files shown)


src/BootstrapTable.js
```diff
@@ -215,6 +215,9 @@ class BootstrapTable extends Component {
     let { replace } = nextProps;
     replace = replace || this.props.replace;
 
+    if (nextProps.data === undefined) {
```
AllenFang: try following instead
if (!nextProps.data) {
  return;
}

athenaxing:
Thank you.




我思故我在
On 07/08/2017 17:58, Allen wrote:

@AllenFang requested changes on this pull request.

In src/BootstrapTable.js:

> @@ -215,6 +215,9 @@ class BootstrapTable extends Component {
     let { replace } = nextProps;
     replace = replace || this.props.replace;
 
+    if (nextProps.data === undefined) {


try following instead
if (!nextProps.data) {
return;
}

src/BootstrapTable.js
athenaxing: Have changed the code to 
if (!nextProps.data) {
  return;
}

AllenFang:
Release on `v3.4.5` thanks your contribution

PR #1464, opened 2017-07-09
Allow setting attributes on editable inputs

jspaine:
I think it would be nice if there was an easy way to set attributes for editable inputs without having to create a customEditor. This allows stuff like:
```
editable={{type: 'number', attrs: {min: 1, max: 4, step: 1}}}
```
I'd only want it for simple attributes, but I don't like that it just silently ignores anything the user passes for ref, onKeyDown and onBlur and I'm not sure the best way to deal with it. Is it worth the extra code to support these, should ref and on* throw an error...?

### DIFF
--- a/src/TableEditColumn.js
+++ b/src/TableEditColumn.js
@@ -169,7 +169,8 @@ class TableEditColumn extends Component {
       row
     } = this.props;
     const { shakeEditor } = this.state;
-    let attr = {
+    const attr = {
+      ...editable.attrs,
       ref: 'inputRef',
       onKeyDown: this.handleKeyPress,
       onBlur: this.handleBlur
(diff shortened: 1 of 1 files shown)


src/TableEditColumn.js
```diff
@@ -179,6 +179,7 @@ class TableEditColumn extends Component {
     let { className } = this.state;
     // put placeholder if exist
     editable.placeholder && (attr.placeholder = editable.placeholder);
+    editable.attrs && (attr = {...editable.attrs, ...attr});
```
AllenFang: I think following will be simple, keep `const` and dont assign `attr` again:

```js
const attr = {
      ref: 'inputRef',
      onKeyDown: this.handleKeyPress,
      onBlur: this.handleBlur,
      ...editable.attrs
};
```

Please check 👍   thanks

jspaine: Yeah that's easier :grin: I spread the user supplied attrs first though so the ref and handlers can't be overwritten.

jspaine:
I added the code to allow refs, keydown and blur handlers. Also a warning 'Setting editable.placeholder is deprecated. Use editable.attrs to set input attributes' so you can eventually get rid of separately setting the placeholder attr.
Does it sound ok? Shall I push it here or do you want to leave it like this?

AllenFang:
@jspaine, it sounds good, feel free to patch PR 👍

AllenFang:
@jspaine, this was fixed and released on `v4.0.0-beta.8`, thanks your contributions 👍

PR #1479, opened 2017-07-13
fix double validation error

jspaine:
Pressing tab in an invalid editable field shows the validation error twice. Added a param to the validator function to disable notification in that case.

### DIFF
--- a/src/TableEditColumn.js
+++ b/src/TableEditColumn.js
@@ -29,7 +29,7 @@ class TableEditColumn extends Component {
                       this._getCheckBoxValue(e) : e.currentTarget.value;
 
       // TAB triggers blur so no need to notify
-      const notify = e.keyCode !== 9;
+      const notify = e.keyCode !== 9 || !this.props.blurToSave;
       if (!this.validator(value, notify)) {
         return;
       }
--- a/src/TableEditColumn.js
+++ b/src/TableEditColumn.js
@@ -24,13 +24,16 @@ class TableEditColumn extends Component {
 
   handleKeyPress = e => {
     if (e.keyCode === 13 || e.keyCode === 9) {
-      // Pressed ENTER
+      // Pressed ENTER or TAB
       const value = e.currentTarget.type === 'checkbox' ?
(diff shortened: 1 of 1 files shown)


AllenFang:
@jspaine thanks your contribution, I'll test it more, and did you enable another options? for example: expand, selection or something?

jspaine:
No, I just tried it with editable number, editable textarea with save button and custom editor.

jspaine:
Just realized I can use the examples to test... It seems to break expand-row-with-cellEdit but I'm not really sure what the expected behavior is. Here's what happens:

| version | blurToSave | input valid | press tab: |
|-----------|-----------------|----------------|-------------------|
| original  | false | true | blurs and appears to save|
| original | false | false | shows validation error |
| original | true   | true | blurs and saves |
| original | true  | false | shows double validation error |
| this       | false | true  | blurs and saves |
| this       | false | false | input shakes but no error shown |
| this       | true  | true | blurs and saves |
| this       | true  | false | shows single validation error |

I didn't see any examples of editable with selection but I'll see if I can test it too.

AllenFang:
ok, so you means this PR will cause something broken for [expand-row-with-cellEdit.js](https://github.com/AllenFang/react-bootstrap-table/blob/v4.0.0-dev/examples/js/expandRow/expand-row-with-cellEdit.js) example?

and for using selection and celledit, I will check it also, BTW, if you want to use them both, there's nothing magic, But if you want to click to edit and also selection row, you need to configure `options.selectRow.clickToSelectAndEditCell` instead of `options.selectRow.clickToSelect`

jspaine:
I think the problem is that blurToSave=false doesn't really work, not because of this PR. I hadn't tried it before, but editing a cell in a table without blurToSave and pressing tab causes it to be saved...

The difference with this PR is that if you change the cell to something invalid and press tab it tries to save but doesn't trigger blur, so the cell shakes but no error is shown. If you press enter it works fine.

A quick fix is to just check if blurToSave is false and show the notification then too:

AllenFang:
Released on `v4.0.0`, thanks your contributions.

PR #1487, opened 2017-07-17
Added prop defaultValue so users can display a default value for empty data

bradleysmagacz:
In some cases, the user may want to display a default value in the table if no data is available instead of a blank value. The user can add a 'defaultValue' prop to the BootstrapTable component if need be.

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -425,6 +425,7 @@ class BootstrapTable extends Component {
             tableBodyClass={ this.props.tableBodyClass }
             style={ { ...style, ...this.props.bodyStyle } }
             data={ this.state.data }
+            defaultValue={ this.props.defaultValue }
             expandComponent={ this.props.expandComponent }
             expandableRow={ this.props.expandableRow }
             expandRowBgColor={ this.props.options.expandRowBgColor }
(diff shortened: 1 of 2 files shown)


AllenFang:
HI @bradleysmagacz so this default valuse will be apply on every columns when `data` is empty array on `BootstrapTable`?

bradleysmagacz:
Hi @AllenFang, this will actually only apply at object key level. For example, lets say I have an array of objects that look like this:

`[{
  name: 'Greg',
  level: 1
}, {
  name: 'John',
  level: 2,
  weight: 180
}]`

And let's say the user passes in 'No Data' as the defaultValue prop. When the table renders, Greg's weight will be set to 'No Data' instead of an empty field. If the user does not pass in a defaultValue prop, everything will function as before.

Although it's not a common use case, there are situations in our application where we need to render a default value when the object key does not exist or no data is available.

AllenFang:
@bradleysmagacz, I see, thanks your descriptions 👍

AllenFang:
HI @bradleysmagacz , after review, I think the `defaultValue` should be depend on every different `TableHeaderColumn`, for your above example, `weight` may have 50 as default, `level` may have 1 as default and so on, do could you please have me to change for this?  If you dont know how to handle it, I can help you, but just let me know, thanks 👍

bradleysmagacz:
Hi @AllenFang. How can the defaultValue prop be passed down to `TableColumn`, where the data value is being rendered, if the prop is being passed into `TableHeaderColumn`? Currently I only see a way to pass down the defaultValue prop to `TableColumn` through `TableBody`.

stephanebarbey:
If I may add something: I'd rather have the default value be a PropTypes.func than a PropTypes.string, so that we can make it work nicely if sorting or filtering is invoked (the value for a cell may be a Date or a number.)

stephanebarbey:
Also: don't you want this default value to be proposed (by default) in the Add Row dialog? (This is what I would understand of a column's default value.)

AllenFang:
@bradleysmagacz 
1. get the defined on [here](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/BootstrapTable.js#L144), this function transfer each props on `TableHeaderColumns` into an plan object.

2. Actually, I already pass the columns object into `TableBody`, [here](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/BootstrapTable.js#L434), so you can get each columns props data in the TableBody.

3. [here](https://github.com/AllenFang/react-bootstrap-table/blob/master/src/TableBody.js#L113) render the `TableColumn`, so you can pass the default value into it.

Let me know if you still have any questions, thanks

PR #1516, opened 2017-07-27
Fix PropTypes for React 16

inDream:


AllenFang:
@inDream thanks your contributions 👍

PR #1555, opened 2017-08-16
Avoid pushing null table row data to store

michaeljayt:


AllenFang:
HI @michaeljayt I wanna know what problem you encounter, thanks 👍

PR #1556, opened 2017-08-16
Add quotation marks to header cells on CSV export

lwatt:
You add quotations to avoid commas in cell values breaking the formatting in the line:

`var cell = typeof value !== 'undefined' ? '"' + value + '"' : '';`

but not in when exporting header cells, meaning that header cells which include commas are split when viewing the CSV file in programs like excel.


AllenFang:
Thanks your contribution 👍

AllenFang:
Released on `v4.0.2` :)

Issue #1568, opened 2017-08-20
placeholder-selected in filters/Select.js is not handled correctly


CroniD:
Hi there,

the css class `placeholder-selected` is not removed from class names in filters/Select.js, when an option is selected not equal to the placeholder. The not-placeholder option is displayed in something like light grey if selected. Formatting in the menu is correct tho, but not in the field of the select element. There is no workaround for this. :(

Best regards

AllenFang:
@CroniD thanks, I'll handle it :)

AllenFang:
Fixed on `v4.0.3`

CroniD:
@AllenFang & @rennerDa Thank you both. It works. :)

PR #1576, opened 2017-08-29
Changed isPlaceholderSelected check in componentWillReceiveProps

rennerDa:
IMO there is no need to check for defaultValue here. Just check for current selected value. Maybe this method can be removed, the state for isPlaceholderSelected is updated on every action in this class (f. e. filter, ...).
This fixes the github-issue #1568 - "placeholder-selected in filters/Select.js is not handled correctly" - (https://github.com/AllenFang/react-bootstrap-table/issues/1568)

### DIFF
--- a/src/filters/Select.js
+++ b/src/filters/Select.js
@@ -23,7 +23,7 @@ class SelectFilter extends Component {
     };
   }
 
-  componentWillReceiveProps(nextProps) {
+  componentWillReceiveProps() {
     const currentSelectValue = this.refs.selectInput.value;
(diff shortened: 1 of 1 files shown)


rennerDa:
This was a fast fix, please check if I forgott some state changes or something else.
The examples are working as expected (but I only testet the filter examples).

AllenFang:
@rennerDa thanks your contribution, I'll check this soon.

AllenFang:
Released on `v4.0.3` 👍

PR #1579, opened 2017-08-30
setting environment to production

hrajotia:
This will reduce output file size and also prevent __REACT_HOT_LOADER__ related stuff getting added to minified file.

File size:
Original: **~251KB**
After change: **~226KB**


AllenFang:
@hrajotia it's awesome 👍

AllenFang:
Release on `v4.0.3`

Issue #1202, opened 2017-04-06
How can i filter on multiple values?


rahul1059:
I am using handleFilterData for custom filters. I have checkboxes in select dropdown, and can select multiple values from dropdown. If i pass mutiple values as array e.g. ["All","onlyOne"] to handleFilterData, its not working. However if i pass only one value it works fine like ["All".

Do you have any solution for this ?

sunnybogawat:
I am also facing the same issue and want to know is there any way to pass an array of values instead of single value.

AllenFang:
ok, I'll try to support it, but It may need some time

jsangilve:
Hi @AllenFang, have you made any progress on this issue? 
I've been reading the code and I believe all changes required to support this feature should be made on `TableDataStore`'s methods? I could help you to do this.

AllenFang:
sorry, I'm busy recently, PR is welcome, feel free to discuss with me if you have any questions.

AllenFang:
BTW, I'll review this issues and try to implement it again in near future.

bpsourav21:
Hello guyz,
Somehow I managed to solve similar problem as you asked.I have used custom filter. And modified this filter as a select tag. In "Select Filter" it has an issue. It always required a object which is designed like below:
option:{0: "Rice", 1:"Pants", 2:"Shirts"}

But I have modified them using custom filter where I can use a list of array.

the demo code is given below:
bootstrap table code:
```

getCustomFilter(filterHandler, options) {
        return (
            <CustomFilter filterHandler={filterHandler} />
        );
    }
  render() {
 return (
<div>
<BootstrapTable data={this.props.all_products} search={true} pagination={true}>
                <TableHeaderColumn width="25%" dataField='product_code' isKey={true}>Product Code</TableHeaderColumn>
                <TableHeaderColumn width="25%" dataField='product_name'>Product Name</TableHeaderColumn>
                <TableHeaderColumn width="15%" dataField={this.state.selectField}>Product Name</TableHeaderColumn>
                <TableHeaderColumn width="15%" dataField="category" filter={{ type: 'CustomFilter', getElement: this.getCustomFilter.bind(this) }}>Category</TableHeaderColumn>
                <TableHeaderColumn width="20%" dataField='createdAt' dataFormat={this.dateFormatter.bind(this)} >Date</TableHeaderColumn>
            </BootstrapTable>
</div>
);



```
CustomFilter.js



    dataSelected(filterHandler, e) {
[...]

PR #1624, opened 2017-09-15
Add new FILTER_TYPE 'ArrayFilter' to allow filtering on multiple values

jsangilve:
This is workaround for issue #1202. It adds a new FILTER_TYPE called `ArrayFilter` for filtering multiple values.

@AllenFang I know you have focus development on react-boostrap-table2, but I think this would be a small but useful improvement.

```
export default class ProgrammaticallyArrayFilter extends React.Component {
  /* Filtering passing an array of values */
  handleBtnClick = () => {
    this.refs.table.handleFilterData({
      name: { type: 'ArrayFilter', value: [ 'Item name 3', 'Item name 4' ] },
      price: { type: 'ArrayFilter', value: [ 2100, 2104 ] }
    });
  }

  render() {
    return (
      <div>
        <button onClick={ this.handleBtnClick } className='btn btn-default'>Click to apply text filter</button>
        <BootstrapTable ref='table' data={ products }>
          <TableHeaderColumn ref='nameCol' dataField='name' filter={ { type: 'TextFilter', delay: 1000 } }>Product Name</TableHeaderColumn>
          <TableHeaderColumn dataField='price'>Product Price</TableHeaderColumn>
        </BootstrapTable>
      </div>
    );
  }
}
```

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -67,7 +67,7 @@ class BootstrapTable extends Component {
       }
       if (column.props.isKey) {
         if (keyField) {
-          throw new Error('Error. Multiple key column be detected in TableHeaderColumn.');
+          throw new Error('Error. Multiple key column detected in TableHeaderColumn.');
         }
         keyField = column.props.dataField;
       }
@@ -82,6 +82,7 @@ class BootstrapTable extends Component {
       }
     });
 
+    // if a column filter was created, add 'onFilterChange' listener
     if (this.filter) {
       this.filter.removeAllListeners('onFilterChange');
       this.filter.on('onFilterChange', (currentFilter) => {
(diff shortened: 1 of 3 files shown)


masoudmanson:
OMG i was looking for this whole day :0
Pleeease merge it ASAP
Thank you for adding this feature guys

CEmocca:
thanks man i'm looking for this features for a while, i try to use custom filter. But it's end up with some kind of error when i try to use it on more than one column.

AllenFang:
@jsangilve nice fix, I'll merge it on next version if without any critical bug 👍

AllenFang:
Hi all, this is released on `v4.0.5`.  [Example](https://github.com/AllenFang/react-bootstrap-table/blob/master/examples/js/column-filter/array-filter-programmatically.js) like @jsangilve mention above. and thanks  @jsangilve  again.

PR #1640, opened 2017-09-22
updated .sort in TableStore.js to handle undefined values

stevelacerda7:
.sort handled null values, but not undefined, which was created some crazy sorting behavior. Thus, I modifeed .sort to handle both null and undefined values, which I tested and seems to work fine.


AllenFang:
@stevelacerda7 sorry for lately checking, this patch is ok for me but could you please help me to remove the commit for the changes under the `dist` folder? Just commit the source code, thanks 👍

stevelacerda7:
Sorry about the git mistakes, new to this, so working through some issues. I created a new pull request, please check it out and let me know if there's anything else that I need to do. Thanks.

Issue #1646, opened 2017-09-26
TypeError: Cannot read property 'emitter' of undefined


widdde:
When upgrading from react bootstrap table v 4.0.0 to v4.0.5 I get 
TypeError: Cannot read property of 'emitter' of undefined
at TableHeaderColumn.componentWillReceiveProps

Fails in:
`
if(nextprops.dataField !== this.props.dataField) {
var emitter = nextprops.filter.emitter || {} < -----
}`

In my table I use the following options:

`<BootstrapTable data={this.state.data} cellEdit={this.cellEditProp} remote={this.remote}
options={{noDataText: "no Data",
onCellEdit=this.onCellEdit.bind(this),
clearSearch: true,
onDeleteRow: this.onDeleteRow.bind(this,
insertModalHeader: this.createCustomModalHeader,
insertModalFooter: this.createModalFooter,
onAddRow: this.onAddRow.bind(this)}}

insertRow={true}
deleteRow={true}
selectRow={{
mode: 'radio',
bgcolor: 'pink',
beforeDeleteRow: this.beforeDeleteRow.bind(this)}}

version='4'
hover search striped>
<TableHeaderColumns..
< /BootstrapTable >
`

I don't get this error in earlier versions of bootstrap table.. 
One of my tables works, but another (the above) with more edit and add options doesn't work at all. 

EDIT: All works OK in v 4.0.3 as well, but not in v.4.0.5.

EDIT2: It fails when I have this simple <TableHeaderColumn:
`<TableHeaderColumn dataField='pattern' tdStyle={{whiteSpace: 'normal', fontWeight: 'lighter' }}>
Validation</TableHeaderColumn>
`
When I remove this tableHeaderColumn the table works fine. This is very strange. I have the exact same setup on other TableHeaderColumns, and they work perfectly..

When I change the order of my TableHeaderColumns it starts to work. How is this possible? 
If I move down this TableHeaderColumn to under another TableHeaderColumn everything works, and when I move it back the table throws TypeError as said above.

If I move the column under this column everything works..
`<TableHeaderColumn datafield='required' dataFormat={this.BooleanFormatter}
dataAlign='center' width='120' thStyle={{ 'fontWeight: 'lighter'}}
editable={{'type': 'checkbox', options: {values: 'true:false'}}}>Is Required</TableHeaderColumn>
`

marcionicolau:
In the first render of data, everything is ok, when the data change, throws the error below.

[...]

PR #1654, opened 2017-09-29
Fix filter undefined

### LINKED ISSUE #1642: When receiving new data, which is different than previously rendered, table errors with emitter undefined.
Similar to [#1418](https://github.com/AllenFang/react-bootstrap-table/issues/1418). I have a table that's being redrawn on a time interval, as data changes, so the columns and data may change as new data comes in. The problem is that the table doesn't want to cooperate when the data structure changes (new columns are added). I tried replace=true on the BootstrapTable, which only seems to work if I use:

```
setState({ rows: [], columns: [] })
setState({ rows: newData, columns: newData: []);
```

Otherwise, I receive an error:

`bundle.js:223729 Uncaught TypeError: Cannot read property 'emitter' of undefined`

The problem is, that I then get two renders, and I can visually see the render of the empty dataset. Is there anyway to handle the new column or removal of an old column gracefully?

inDream:
Fix #1642 #1646 for 84cc76c.

### DIFF
--- a/src/TableHeaderColumn.js
+++ b/src/TableHeaderColumn.js
@@ -24,7 +24,7 @@ class TableHeaderColumn extends Component {
     }
 
     // If column not displaying the same dataField, reset the filter accordingly
-    if (nextProps.dataField !== this.props.dataField) {
+    if (nextProps.filter && nextProps.dataField !== this.props.dataField) {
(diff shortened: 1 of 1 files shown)


AllenFang:
Release on `v4.0.6`, thanks your contribution 👍

PR #1726, opened 2017-10-31
Add fetchInfo into doc

NumminorihSF:
Current on-line doc does not include any info about `fetchInfo` property, and it's very confused.


AllenFang:
@NumminorihSF thanks your contribution

PR #1742, opened 2017-11-08
Nested data fields

JcBernack:
Added the ability to access nested properties.

My solution uses lodash.get and set to read/write from the row object. The dataField property is used as the accessor, e.g. `dataField='user.foo.bar[1]'` will access `row.foo.bar[1]`.

The only exception is the `keyField` which can not be a nested property.


fengshuo:
I was thinking the same thing and found this pull request 👍 
does this pr solve this case:
```
data=[
  {
     a: 50,
     someData: {
        dataA: 100,
        dataB: 100
     }
  }
]
and use it like this?
dataField='someData.dataA'
```

JcBernack:
Yes, exactly. I can not guarantee that it works in all cases though. If you check it out please let me know how it went.

AllenFang:
Hey all, firstly thanks your contributions, actually, I'm already stop all new feature imeplement and start to move to [`react-bootstrap-table2`](https://github.com/react-bootstrap-table/react-bootstrap-table2/tree/develop) and we already support nested data but not released yet.

however, this PR is good but still missing some changes and it's not very easy to implement nested data in current code base, so I consider not to merge this PR and will encourage people to use `react-bootstrapt-table2` in the future.

Let me know your any thinking or idea and very appreciated your contributions, thanks

Issue #1752, opened 2017-11-14
Sorting field and order is not taken from props when using remote datasource


aalbul:
I found that when i use remote datasource, there is no way to reset / change sortName or sortOrder.
My use case is pretty easy. I store table data, pagination data + sorting data in redux. When i change sorting field in Redux (and not by clicking on header), this change is not reflected in the table.

This is happening because this data is not synchronised back into store inside of BootstrapTable.componentWillReceiveProps method.

aalbul:
I created pull request to fix it. Please review it here: https://github.com/AllenFang/react-bootstrap-table/pull/1753

AllenFang:
@aalbul I'll check this out, thanks

AllenFang:
@aalbul I've some feedbacks on PR

aalbul:
Thanks, please see my answer in PR :)

AllenFang:
@aalbul released on `v4.1.5`. thanks your contributions

dawnmist:
Prior to this change, clearing a sort used to work by:
* Using Table.cleanSort(), and
* Setting `SortOrder = undefined`, `SortName = ''` in the table options.

Doing this with 4.2.0 (I would assume in 4.1.5 - I just tried upgrading from 4.1.4) results in BootstrapTable throwing the error: `"The type of sort field and order should be both with String or Array"`.

The changes made for this issue do not take into account/provide any way to reset the current sort without triggering the above error check. i.e. `TableDataStore.setSortInfo(sortOrder, options.sortName)` enforces that sortOrder must never be undefined.

PR #1753, opened 2017-11-14
#1752 Sorting field and order is not taken from props when using remo…

aalbul:
…te datasource

### DIFF
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -283,6 +283,12 @@ class BootstrapTable extends Component {
           }
           newState.data = data;
         }
+
+        const sortName = options.sortName;
+        const sortOrder = options.sortOrder;
+        if (this.props.options.sortName !== sortName || this.props.options.sortOrder !== sortOrder) {
+          this.store.setSortInfo(sortOrder, options.sortName);
+        }
         this.setState(() => newState);
       } else {
         // #125
--- a/src/BootstrapTable.js
+++ b/src/BootstrapTable.js
@@ -284,9 +284,10 @@ class BootstrapTable extends Component {
           newState.data = data;
         }
 
+        const { options: currentOptions } = this.props;
         const sortName = options.sortName;
         const sortOrder = options.sortOrder;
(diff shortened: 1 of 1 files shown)


src/BootstrapTable.js
```diff
@@ -283,6 +283,13 @@ class BootstrapTable extends Component {
           }
           newState.data = data;
         }
+
+        const { options: currentOptions } = this.props;
+        const sortName = options.sortName;
+        const sortOrder = options.sortOrder;
+        if (this.allowRemote(Const.REMOTE_SORT) && (currentOptions.sortName !== sortName || currentOptions.sortOrder !== sortOrder)) {
+          this.store.setSortInfo(sortOrder, options.sortName);
+        }
```
AllenFang: This part only set the sort field and order to store object. but the data still not change by newest sort field and order.

My suggestion is if you change the sort field or order programmatically(not by clicking on header), then you are suppose to handle the sorting function in your program and then feeds newest data(sorted) into table

Because the remote mode will not handle the data change(sort, filter, pagination) internally.

aalbul: That is exactly my intention. The data is sorted on the backend and passed into the table via props. Sort order field(s) changed programatically sometimes (assume case when you need to reset the page to defaults).

So in other words, i assume that if table works in "remote" mode, then all the data logic is handled "remotely" and component is dumb (or almost dumb) and only renders stuff that is passed to it. This fix making sure that if you change sort order from outside, it get's synchronized back to internal variable inside of the table and renders properly.

AllenFang: @aalbul got it, sorry I got some misunderstand before, but now is clear for me ;) thanks

AllenFang:
@aalbul I merged your commit manually https://github.com/AllenFang/react-bootstrap-table/commit/9dacb793b91ca498e6a6a4417bdb4d510b31093b and https://github.com/AllenFang/react-bootstrap-table/commit/06b356021e3364d0048f8aba8e934e07bb7a02fc, hence close this PR.

Thanks your contributions

PR #1754, opened 2017-11-14
Wrapping header content in a div element

michal-lipski:
Wrapping header content in a div element in order to apply styles on all of the header elements without modifying styles of the header.

The real world case for me was to move caret (sorting indicator) before header text. The easiest way is to use display flex and flex-direction: row-reverse. Problem is that I cannot apply this styling on header directly, as it needs to have display table-cell to work properly. With this wrapper, it will be easier to style content of the header.

### DIFF
--- a/src/TableHeaderColumn.js
+++ b/src/TableHeaderColumn.js
@@ -181,9 +181,11 @@ class TableHeaderColumn extends Component {
           colSpan={ this.props.colSpan }
           data-is-only-head={ this.props.isOnlyHead }
           { ...attr }>
-        { children }{ sortCaret }
-        <div onClick={ e => e.stopPropagation() }>
-          { this.props.filter && !isOnlyHead ? this.getFilters() : null }
+        <div className='header-content'>
+          { children }{ sortCaret }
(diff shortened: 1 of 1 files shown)


AllenFang:
Sorry for lately reply,

HI @michal-lipski I'm not sure if following will help you:
```<TableHeaderColumn dataField='name' headerText='Product Name'><div>Product Name</div></TableHeaderColumn>```

the `children` will wrap a custom div, but not sure if it's will be more easy for you to move the sort caret.

If it doesn't help, I think your PR is ok, but I hope we can have a prop to control if wrap a div or not

Let me know you idea, thanks your contributions.

michal-lipski:
Thanks, but this will not help as carret componenet is rendered outside of a childern: `{ children }{ sortCaret }`

Ok, I see that wrapping in a div might be a breaking change. Someone could rely their style on direct children of <th> element.

How about `wrapHeader` property? Or `wrapHeaderContent` property? By default, it will be equal to false.

AllenFang:
@michal-lipski  how about `wrapHeaerCell`? and agree the default is `false`

PR #1778, opened 2017-11-27
Expand hide row

powellandy:
hide the original row when expanded.

This is useful when the expanded row is an edit form.  We have a lot of edit rules in place, and having a form on a row makes the UI flow better.  Hiding the original row keeps it cleaner.

I am trying to only include the changes for the react components, but somehow i have a the dist changes in here too.  The source will need to be rebuilt as the dist changes in this PR are not built from the source of the PR files.


powellandy:
I have the dist changes removed.  Please let me know if there is anything else.  Thanks and thanks for this project, it has been very useful!

AllenFang:
Released on `v4.3.0`. Thanks you 👍

AllenFang:
https://github.com/AllenFang/react-bootstrap-table/commit/c54c54bd0a927788d2eeeb25613e82c4c77daeb2

PR #1786, opened 2017-12-02
fix issue with key when using custom page elements

### LINKED ISSUE #1785: Custom pagination element doesn't work
Hi, when I try to use custom elements for pagination, it doesn't work. Eg:

```
const options = {
  prePage: <i className="test1" />,
  nextPage: <i className="test2" />,
  firstPage: <i className="test3" />,
  lastPage: <i className="test4" />
};
```

Reproducer:
https://codesandbox.io/s/ymp0plo6r9

VictorChen:
For the case where `prePage`, `nextPage`, `firstPage` and `lastPage` are custom elements, the pagination breaks because we're passing them as the `key` (which turns them into `[object Object]` and we end up with duplicate keys). A simple fix is to just use the page index as the key.

Fixes #1785 

Also, fix react proptype warnings for`prePage` (it shouldn't be a `string`) and add proptypes for the other

### DIFF
--- a/src/pagination/PaginationList.js
+++ b/src/pagination/PaginationList.js
@@ -223,7 +223,7 @@ class PaginationList extends Component {
           false :
           true;
       }, this)
-      .map(function(page) {
+      .map(function(page, index) {
         const isActive = page === this.props.currPage;
         const isDisabled = (isStart(page, this.props) || isEnd(page, this.props)) ?
           true :
@@ -241,7 +241,7 @@ class PaginationList extends Component {
         }
 
         return (
-          <PageButton key={ page }
+          <PageButton key={ index }
             title={ title }
             changePage={ this.changePage }
             active={ isActive }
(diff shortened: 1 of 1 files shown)


VictorChen:
hi @AllenFang, any chance we can release a new version with this fix?

src/pagination/PaginationList.js
```diff
@@ -241,7 +241,7 @@ class PaginationList extends Component {
         }
 
         return (
-          <PageButton key={ page }
+          <PageButton key={ index }
```
madeinfree: ref: https://reactjs.org/tutorial/tutorial.html#keys

The tips from reactjs doc, I suggested to use like `${title}-${index}` to instead it ~~

VictorChen: hmm.. would this make a difference since we're including the `index` anyways?

AllenFang: I fixed it

AllenFang:
Released on `v4.3.0` Thanks you guys.

PR #1806, opened 2017-12-12
fix bug with using custom pagination components

### LINKED ISSUE #1785: Custom pagination element doesn't work

VictorChen:
Looks like custom page components are still not working as expected (my previous PR only fixed part of the issue, oops 😅). The problem now is that when you click on the `PageButton`, it calls:

```js
this.props.changePage(e.currentTarget.textContent);
```

It's expecting to get the page number from `e.currentTarget.textContent` which obviously won't work for custom components. Instead, I pass through the page number as a prop to the `PageButton`.

Also added a demo for further testing. I believe everything should work now!

Fixes #1785

### DIFF
--- a/examples/js/pagination/demo.js
+++ b/examples/js/pagination/demo.js
@@ -1,6 +1,7 @@
 /* eslint max-len: 0 */
 import React from 'react';
 import DefaultPaginationTable from './default-pagination-table';
+import IconPaginationTable from './icon-pagination-table';
 import CustomPaginationTable from './custom-pagination-table';
 import PaginationHookTable from './pagination-hook-table';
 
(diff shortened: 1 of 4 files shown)


AllenFang:
@VictorChen coool, I'll check later

AllenFang:
@VictorChen Thanks your contribution, fixed on `v4.3.1`, thanks

PR #1838, opened 2018-01-05
Fixed link to scroll-table.js from the demo page

chugreevd:
Launched the demo locally as described here https://github.com/AllenFang/react-bootstrap-table, the current link led to 404 on github


AllenFang:
@chugreevd thanks your contribution :)

PR #1893, opened 2018-02-10
Fixed an issue with the keyboard navigation when there are hidden columns

DmitryMK:
Hi @AllenFang, thank you for the amazing library, I love it! 

It looks like there is an issue with keyboard navigation in 4.3.1 when there are hidden columns. To replicate it you can create a table with any of the columns hidden and keyboard navigation enabled. It looks like you planned for it by adding the visibleColumnSize variable, but  variable 'x' still contains absolute column number.
What I did, is added an visibleColumnIndices array with indices of the non-hidden columns and made 'x' take only values from it.


AllenFang:
@DmitryMK thanks, I will check it our after my vocation 👍

PR #2056, opened 2019-02-07
Fix: Add separator line at the beginning of created csv

BernatWozzo:
This fix allows Excel to detect the separator when opening the file and fixing the problem of all of the data shown in the same column.
With this change Excel opens csv like an xls without the need to import the csv data manually and specifying the separator.


Ajitprakash:
@AllenFang did you merge this in master ?? @BernatWozzo

